PR #232, opened 2015-01-08
Mac address fields

scarito:
I added source and destination MAC fields to the output field list.  I understand they may be of limited use to the project but they can be helpful when scanning a local network.  If a scanning mode is selected which does not provide the ethernet header, the dummy values used for the probe modules will appear instead of the true MAC addresses.


scarito:
I've realized that even on the local network zmap will throw all its packets at the default gateway.  If that behavior changes in the future it may become more useful.  I'm using this with an ARP scanning module I'm writing and it works correctly for that purpose.

scarito:
Merged into #236

PR #235, opened 2015-01-09
Arp scan

scarito:
I wanted to be able to do a local network liveness scan, so I implemented an ARP module.  This module works a little differently because the probe module framework was designed around receiving IP datagrams.  On the transmit side there are no issues -- the packets are addressed to the broadcast address instead of the gateway, as in a normal ARP request.  On the receive side, zmap tries to decode the ARP packet as an IP header, and thus populates the saddr and daddr fields with bogus values.  The ARP module processes the message and then populates the arp-saddr and arp-daddr fields in order to avoid creating a conflict in the namespace.

I'd like to resolve this in a way that is more consistent (that is, to get the correct values into saddr and daddr).  My ideas for ways to proceed are:
- Add a field to probe_module_t to indicate whether or not the packet should be processed as an IP frame in recv.c (and thus call fs_add_ip_fields)
- Move the responsibility of calling fs_add_ip_fields into the respective probe modules, making probe modules agnostic to the specific network protocol (I think this would have the biggest consequences, since it requires a change in the behavior of existing probe modules).
- Overwrite the values set by fs_add_ip_fields with the correct source and destination (this seems the most hacky to me)

I would also suggest changing the signature of the validate_packet function to pass a raw packet rather than an IP frame simply because that is more consistent with process_packet.

I'd like to find a good way to integrate this, even if it's only useful on the local network, so I'd appreciate suggestions from the devs on how best to proceed.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -53,6 +53,7 @@ SET(PROBE_MODULE_SOURCES
 	probe_modules/packet.c
 	probe_modules/probe_modules.c
 	probe_modules/module_ntp.c
+	probe_modules/module_arp.c
 	)
 
 SET(SOURCES
--- a/src/probe_modules/module_arp.c
+++ b/src/probe_modules/module_arp.c
@@ -1,0 +1,191 @@
+/*
+ * ZMap Copyright 2013 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+// probe module for performing ARP scans
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <time.h>
+#include <unistd.h>
+#include <string.h>
+
+#include "../../lib/includes.h"
+#include "probe_modules.h"
+#include "../fieldset.h"
+#include "packet.h"
+#include "validate.h"
+
+probe_module_t module_arp;
+
+#define ARPHRD_ETHER 1
+
+#define ARPOP_REQUEST 1
+#define ARPOP_REPLY 2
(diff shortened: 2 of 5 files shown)


zakird:
My personal preference is go move the call to add IP fields to each probe
module and not add an additional flag. Seems a bit cleaner.

As to validate packet, I have no objection to passing entire Ethernet frame.

zakird:
Also, for the record, I'm totally happy to integrate this into zmap once
we've gotten those couple of issues cleared up.

Cheers
Zakir

dadrian:
As for the pull request, I'm also happy to merge once we clean up the issues. I could see either modifying the probe_module_t or simply modifying existing probe modules to call fs_add_ip_fields(). Conveniently, we're in the process of a major release (2.0.0 from 1.2.2), so we don't have to worry about breaking existing APIs.

scarito:
I can go ahead and change the signatures and move the call to fs_ip_add_fields if you would like.  I'll make a separate branch/PR for those changes.

As far as the make_mac_str function I didn't see another one.  The closest I found was print_macaddr in packet.c which takes a struct ifreq*.  If there is another function I'm missing I'm happy to change that.

Michael

dadrian:
I was probably thinking of print_macaddr, so just leave make_mac_str.

Does the OS not provide headers for ARP packets, or is there another reason you chose to define your own structs? I haven't done much programming around ARP myself, so I'm not familiar with what API Linux or BSD provide.

scarito:
The ARP defines in the beginning as well as the arphdr struct are from net/if_arp.h.  There isn't a struct I've found which includes the IP and MAC addresses after the header.  To be honest the reason I did that is because I'm also cross compiling to ARM and the target sysroot has different header files.  I had some similar troubles getting the rest of zmap to compile because of differences between the Linux and BSD tcphdr definition.  In this case I suppose I could include the header and get rid of those lines.

scarito:
Merged into #236

PR #236, opened 2015-01-10
Probe module refactor

scarito:
This is a first step towards moving the IP related code inside the probe modules.  I also tried to make the various probe module related functions (print, validate, and process) use the same form of the packet.  Let me know if this looks like it's going in the right direction.

The one issue I see is that the list of available output fields is going to be incorrect because it will always include the IP fields (in zmap.c:492).  If we don't want to special case things, one resolution I was thinking of is to have a reference to a "base" module in the probe module definition.  This might be excessively modular but for example I could see the tcp_synscan module referring to an IP module which would refer to an ethernet module.  This may be too complex for the goals of the probe module framework.  Otherwise there just needs to be a way of figuring out whether or not to include the IP fields.


dadrian:
Changes so far look good to me. As far as adding the relevant fields, one approach might be to switch the `	fielddef_t *fields;` field in the `probe_module_t` struct to be an array of field arrays (e.g. `fieldef_t **fields`, and then have the ZMap framework union all the fields together, e.g.

```
probe_module_t tcp_module = {
...
    .fields = {
        IP_FIELDS,
        TCP_FIELDS
    },
...
};

fielddef_t *TCP_FIELDS = {
   ...
};
```

@zakird, thoughts?

scarito:
I've made those changes to the existing modules and the ones that were added recently.  I didn't change the signatures of the two new modules to match but that could be done quickly.  I'm using the fielddefset_t struct in order to hold the length of the various field entries.  Is there anything else you think should be changed to properly integrate this?

scarito:
I've synced up with the master again and everything seems to be working for me.  If anyone else can test this and get it merged I can bring the other two pull requests up to date (since they will depend on this code).  Thanks!

dadrian:
We'll look at getting this merged on @amirian28 finishes ironing out a few bugs in the current implementation of the DNS probe module.

scarito:
I've merged the current master into this PR and tested all of the modules.  I don't see any current issues.  I also went ahead and merged my other two PRs into this one because the functionality is all related.

dadrian:
@scarito: First off, I'm incredibly sorry about the delay on this issue. The ZMap team has been barely able to tread water for the last five months. Do you want to continue ahead with this process? Now is the time to do it, since we're working on switching over from v1 to v2, we get to break APIs :).

It looks like a lot of stuff has changed since the initial pull request. I'm not incredibly familiar with all the changes, do you have a sense on how hard this would be to merge against the current master? Have you used this module much, and do you have a sense of whether we made the right design decisions along the way?

scarito:
No worries.  I'd be happy to help you guys get the stuff ported over.  At this point I'm not sure if the best course of action is to break apart that PR again or to merge and update (which I think I could do fairly quickly).  I haven't used the code much recently but we were using it quite a bit a few months ago and it seemed effective for our purposes.

The things I had to change along the way were mostly related to sending non-IP packets on a local network (even though ARP is still IP-oriented).  I think there were some inconsistencies between the decode, identification, and send and receive functions for plugins.  I tried to normalize that but it's possible a better job could be done.

What do you think would be the best way to proceed?
________________________________

dadrian:
Probably the best thing to do is update this pull request to work with latest master, and then we'll move forward from there.

scarito:
It wasn't too difficult to merge master back in to this branch.  Is there a test suite that can be run to verify these changes?  I'm concerned that I can't effectively test them and this change modifies every probe module.  Particularly the change from IP header to ethernet header and the cascading calls in the _process_packet function need to be verified correct.

I think there's more minor restructuring that could be done (and I'd be happy to help once we get this dealt with) but overall this should make the module interface more consistent and flexible.

dadrian:
Unfortunately, we don't have a comprehensive test suite. Most of the testing we do is manual. I think what we'll likely do is merge this in after we release 2.1 final (which should be within the next week or two), and then test it along the way to a version 2.2 release.

scarito:
@dadrian Hey, anything else you'd like to see to get this closed out?

dadrian:
@zakird, have we merged the necessary changes from master into stable for 2.1? I'd like to start the process of merging this in, but it'll make both of our lives a lot easier if we close out 2.1 first.

zakird:
I'm fine with tagging 1.2.0 at this time.

dadrian:
OK, so we finally released 2.1.0, and then a quick follow-up patch release 2.1.1. @zakird has some probe module and fieldset changes coming in for 2.2.0 that help a lot with the DNS module. Once we get that in (should be in the next week or so), I'll start pulling these changes into master.

We'll target this and the DNS / fieldset changes for 2.2.0. If we lag behind, we can always bump this to 2.3.0 and push out DNS as 2.2.0.

I don't plan on doing anymore point releases for the 2.1.* cycle, unless we find a critical issue, so we don't need to put in too much effort into making sure we keep changes contained enough to cherry-pick over to 2.1.1 (although, obviously the more contained the commits the better).

Thanks for all your patience on this, sorry it's taking so long. The ZMap team has pretty much been underwater since last December.

dadrian:
I've started looking through this. The files I've been through look good so far. I haven't done much actual testing of the executable though.

We'll do some scans with this build on our end, and see if everything still looks good.

zakird:
This has been open for nearly six months now. @dadrian What's going on? If you have issues with the PR, can you please document these? I'd like to see this PR resolved one way or another.

dadrian:
Hmmmm.

We don't have a good way to test this, and in retrospect I'm not convinced this is a good direction to head down, anyway. As it is, we validate packets in too many places. It'd be nice to be able to do ARP scans, but realistically I'm never going to get this to the point where I'm comfortable enough to merge.

Therefore, I'm going to close this. 😢

dadrian:
Sorry for dragging this out for so long. If we're ever in the same city, I'll buy you a beer.

Issue #245, opened 2015-02-10
pcap filter for udp captures all outgoing packets


dadrian:
filter is "udp || icmp", which means every packet we make will get captured by the pcap

dadrian:
@amirian28 and @hmoore-r7, this is now fixed, correct?

hmoore-r7:
Partially. Still getting some unrelated responses even with the current code, I think it narrowed it down to the other port check not working as expected (ran scans with a specific source port and still got replies not destined to our source port). It works OK for now, but due to our odd use case, we have to post-process results now. Fixing the second port check would go a long way to reducing the need to post-process here.

hmoore-r7:
Woops, we definitely fixed the "outgoing treated as incoming" issue (this ticket). There are some additional issues that are causing unrelated replies to get recorded still.

dadrian:
@hmoore-r7 so the problem is now there are valid responses that come in, but are destined to a different port than we were using for outgoing traffic?

hmoore-r7:
Some of the probes we run result in buggy devices spamming the server indefinitely, as a result we get a constant stream of unrelated (to the current scan) UDP traffic. There appears to bug in the dst port checking function (not the src, which we moved already), that is including these unrelated replies, even when using a fixed source port. In other words, send as probe from A to B, we get back a response from C to A on a totally definitely port, and the existing filter seems to have a bug that accepts this.

PR #246, opened 2015-02-17
Drop TTL=255 traffic to avoid logging our own outbound packets.

hmoore-r7:
A small fraction of zmap's outbound traffic is being processed as if it was inbound. A recent SIP scan has ~1000 entries where the traffic is our own and not a probe response. An easy fix is to drop all inbound traffic with a TTL of 255. This fix may be useful in other validation functions as well.

### DIFF
--- a/src/probe_modules/module_udp.c
+++ b/src/probe_modules/module_udp.c
@@ -299,7 +299,7 @@ void udp_print_packet(FILE *fp, void* packet)
 {
 	struct ether_header *ethh = (struct ether_header *) packet;
 	struct ip *iph = (struct ip *) &ethh[1];
-    struct udphdr *udph = (struct udphdr*)(&iph[1]); 
+    struct udphdr *udph = (struct udphdr*)(&iph[1]);
 	fprintf(fp, "udp { source: %u | dest: %u | checksum: %u }\n",
 		ntohs(udph->uh_sport),
 		ntohs(udph->uh_dport),
@@ -382,6 +382,12 @@ int udp_validate_packet(const struct ip *ip_hdr, uint32_t len,
 		__attribute__((unused))uint32_t *src_ip, uint32_t *validation)
 {
 	uint16_t dport, sport;
+
+	if (ip_hdr->ip_ttl == 255) {
+		// drop any of outbound traffic that happened to be caught
+		return 0;
+	}
+
 	if (ip_hdr->ip_p == IPPROTO_UDP) {
 		if ((4*ip_hdr->ip_hl + sizeof(struct udphdr)) > len) {
 			// buffer not large enough to contain expected udp header


hmoore-r7:
Examples of the bad traffic (TTL=255):
```
$ curl -s https://scans.io/data/rapid7/sonar.udp/20150209-dns-53.csv.gz | zcat | head -n 3
timestamp-ts,saddr,sport,daddr,dport,ipid,ttl,data
1423566827,71.6.216.41,35384,149.18.52.82,53,54321,255,34ef010000010000000000000756455253494f4e0442494e440000100003
1423566827,71.6.216.59,41435,37.209.196.220,53,54321,255,34ef010000010000000000000756455253494f4e0442494e440000100003
```

dadrian:
I actually think all ZMap traffic is being processed as if it was inbound in all the UDP probe modules. @zakird and I committed hacked-together validation fixes for UDP, NTP, and UPnP modules based on source IP within the last week or so.

I would like to solve this with pcap. I'm not sure if TTL is the way to go---if there's still a chance we get a packet with TTL 255 we'll still need to implement IP-based validation in the validation function (although it should get ran less often). Is there a way to bind pcap to outgoing/incoming? Most modern ethernet cards have some mechanism for exposing receive queues to userland.

hmoore-r7:
@dadrian Thanks! It looks like TTL isn't the long-term fix for other reasons as well. Zmap is capturing background DNS lookups:

```
$ curl -s https://scans.io/data/rapid7/sonar.udp/20150209-netbios-137.csv.gz | zcat | grep ,8.8.8.8,
 1423561613,8.8.8.8,53,71.6.216.47,42604,63968,49,e5d88082000100000000000020434b4141414141414141414141414141414141414141414141414141414141410000210001
```

One approach might be to limit the source ports to a narrow range (modula destination port number or something). We have an issue with previous scans triggering infinite replies in some cases as well (NATPMP and WDBRPC).

hmoore-r7:
As for PCAP rules for outgoing, the easy fix is to add a BPF rule to limit any traffic with your own MAC address as the source.

dadrian:
MAC address could work, I don't think we do anything funky with it in the ZMap subsystem. It'd be nice to make sure the filter operates in hardware. I imagine MAC address would be, but I'm not sure if there's a way to detect that or not.

I'm also linking this PR to issue #245 for documentation purposes.

hmoore-r7:
Should I close this PR for now? Not sure whether mac filtering happens in hardware or not, but it should be fast enough not to matter much (at least for <10GbE).

dadrian:
Mac filtering sounds like a good approach for now. It won't solve our background UDP noise problem though.

As for 10Gig, we're actually able to open the receive buffer directly when we use PF_RING, so this is less of an issue there.

zakird:
I don't quite understand why our port validation isn't solvign this problem
anymore.

hmoore-r7:
Closing this PR. I will refactor as a pcap filter to only accept traffic from the detected GW address. That approach *might* impact load balancing setups that rely on ARP-based link balancing, but this is pretty rare, and may break with the current GW send code anyways.

hmoore-r7:
@zakird The current pcap filter is to "udp || icmp" and validation was done in two steps. The first step was to check whether the source port of the UDP packet (or the UDP header inside of the ICMP packet) matched our target address. This now only checks in the case of ICMP, because checking for the source port ends up dropping valid traffic that responds from another port back to Zmap. The second check is to make sure the packet matches the port range used by Zmap to send the packet in the first place. This range is wide enough that it ends up including ephemeral ports for unrelated UDP traffic on the same host (DNS, NTP, etc).

hmoore-r7:
Refactored in #248 by prefixing a MAC filter to the BPF.

PR #248, opened 2015-02-27
Ignore traffic that doesn't come from the gateway

hmoore-r7:
This is a retake of issue #246 that prefixes the pcap filter with a rule dropping all traffic that does not originate from the detected or configured gateway MAC address. This seems to comprehensively fix the issue with the receive functions seeing sent traffic.

### DIFF
--- a/src/recv-pcap.c
+++ b/src/recv-pcap.c
@@ -52,7 +52,7 @@ void recv_init()
 	}
 	struct bpf_program bpf;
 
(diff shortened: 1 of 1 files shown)


hmoore-r7:
This still might allow for background traffic, since the source ports used to send UDP probes could be used by the OS at some point for other communication. The fix for that case would be for zmap to actually bind a socket to each of those ports for the life of the scan or to explicitly use a non-ephemeral range by reading /proc.

PR #253, opened 2015-03-17
Change send_run to better handle small rates.

jonasbb:
I want to use zmap but with small rates (of a few packets per second). So I needed to fix the send_loop.
If the packet rate is smaller than the number of thread the interval gets rounded to 0 and is ignored than. The second problem is that the busy loop consumes too much CPU. So I use nanosleep to wait the expected time. This is an additional code path in send_run but shouldn't not matter performance wise due to branch prediction. Also packet rates smaller than number of threads are now possible.

I left the rest of the code unchanged as much as possible because I am not sure if something relies on the divide to 0 behavior.

### DIFF
--- a/src/send.c
+++ b/src/send.c
@@ -213,14 +213,26 @@ int send_run(sock_t st, shard_t *s)
 	int interval = 0;
 	uint32_t max_targets = s->state.max_targets;
 	volatile int vi;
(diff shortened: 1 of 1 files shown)


dadrian:
Thanks!

PR #268, opened 2015-06-08
RIPv1 - UDP module added to examples

chadillac:
A packet template that will trigger responses from RIPv1 enabled routers/devices on UDP port 520.


zakird:
thanks.

PR #282, opened 2015-09-04
Version bump for 2.1.0

cicku:


dadrian:
Thanks for the edits. That being said, I don't won't to move the 2.1.0 tag to this commit, otherwise the Homebrew folks will be mad at us, as well as anyone who fetched the tag already.

cicku:
Haven't used homebrew for years...... Thanks for the tip.

PR #285, opened 2015-09-05
Mass overhaul of ZMap CMake and documentation files.

cicku:
This patch:

* Markdownize the CHANGELOG, BUGFIX first in every release, others go after always.
* CMake syntax cleanup, lowercase for set(), include().
* Introduce CMP0048.
* Refine check for configuration files() to not get overridden.

---

What can be done to do better:

* Please check if @ZMAP_VERSION@ is needed elsewhere.
* Haven't tested on Debian, not sure if it builds still?


dadrian:
@cicku, we want to maintain compatibility with CMake 2.8.12.2, which is the latest CMake version available in the Ubuntu 14.04 LTS package repositories. 

it looks like CMP0048 wasn't introduced until CMake 3.0. I'm not sure if that is a major issue or not, provided we set versions sanely. Unfortunately, the `project()` command looks to be different between 2.8.12 and 3.0, see http://www.cmake.org/cmake/help/v2.8.12/cmake.html#command:project. As a result, CMake is erroring out when I try to build on Ubuntu 14.04.

The `@ZMAP_VERSION@` macro is awesome, especially since I'm so bad at finding all the version strings to update on release. I think we'll need `@ZMAP_VERSION@` in all the man files (`zmap.1`, `ztee.1`, `zblacklist.1`) as well.

Once we get the final build kinks resolves, I'll backport them to 2.1.0 and release a 2.1.1. That should make all the packagers lives easier.

Thanks for the help!

cicku:
I understand, that's why I asked if you can compile on Debian :hamster: 

I need to build it on elder systems as well ( RHEL7 has 2.8.11 only but this version is buggy, even behind RHEL6 which has 2.8.12 already! [See this bug if you want](https://bugzilla.redhat.com/show_bug.cgi?id=1195919).)

So, I think my PR needs some rework:

1. Revert CMP0048, but set(ZMAP_VERSION) manually with 2.1.0, I think this is the compromise.

2. For those manpage, I'm not sure how did you generate them. Are they just the sources, or you used some other tools to build instead? Because I checked and found traces of ronn, I'm afraid not using that macro if it breaks more.

dadrian:
For the manpages, I write them using ronn and then manually run ronn to generate the roff, which I then tweak manually to have the correct header.

Other thing of note, by putting the @ZMAP_VERSION@ in the `*.ggo` files, they always show up as modified in git. We might be able to get around this by having CMake copy the `.ggo` file to some other file that is .gitignore'd before running `configure_file`. Thoughts?

dadrian:
I just confirmed this PR now works with Ubuntu 14.04.

cicku:
@dadrian,

> Other thing of note, by putting the @ZMAP_VERSION@ in the *.ggo files, they always show up as
> modified in git. We might be able to get around this by having CMake copy the .ggo file to some
> other file that is .gitignore'd before running configure_file. Thoughts?

I think you were building it in source, right? Now we have:

`configure_file(topt.ggo ${CMAKE_BINARY_DIR}/src/topt.ggo @ONLY)`, my experience is:

1. Append .in suffix to gengetopt file, and possibly other sources which you'd like to use the macro. (Just do like autoconf oldschool)

2. Change
`configure_file(topt.ggo ${CMAKE_BINARY_DIR}/src/topt.ggo @ONLY)`
to
`configure_file(topt.ggo.in ${CMAKE_BINARY_DIR}/src/topt.ggo @ONLY)`

3. Add `*.ggo` to `.gitignore` file

Done :100:

dadrian:
Sounds good to me

cicku:
Pushed as 4d0d3ac4083dba39481cbacc360b8047b1dc1dbc, I have to sleep now(UTC+8). Any questions please leave here, I will rpely tomorrow(oops, 5 hrs later probably :laughing: )

cicku:
I'm thinking about using cmake to build manpages as well. It should be easier for version bump.

Feasible?

dadrian:
I don't want to require everyone who builds zmap to have to have ronn installed.

cicku:
I know, solutions:

1. You generate manpage via ronn, replace `zmap v2.1.0` with `zmap @ZMAP_VERSION@.` or just `@ZMAP_VERSION@.`  in ronn command. Then include generated .1 file in src/.
2. Use asciidoc (python required, and most of systems have it).
3. Use sphinx manpage builder (python required).
4. Use doxygen (python required, might has timestamp issue).
5. No version in manpage

dadrian:
I think we should probably generate a .1.in file with ronn, use @ZMAP_VERSION@ in that file, and then have CMake generate the .1 file from the .1.in. We'll add .1 file to the .gitignore, and commit the .1.in file to the repository. 

Last time I checked ronn didn't set version in output, so we have to manually edit the ronn output. This workflow is pretty close to what I do now, but should help my general inability to bump version numbers correctly.

cicku:
I used to set version by --organization option of ronn.

dadrian:
Well, if that works, I guess we can just pass @ZMAP_VERSION@ to ronn then. The general workflow should still apply though, with the ronn output being source controlled.

cicku:
Yes, but I haven't had tested this, since ronn is not installed on my system.

dadrian:
OK, I think we're good to merge then. I'll do one last test of this later tonight, then merge to the 2.1.0 branch, bump the version to 2.1.1, release that, then merge this commit into master. We can skip releasing 2.1.0 to package repositories.

cicku:
Good to hear this.

Appreciated!

dadrian:
@cicku, I just merged this to the 2.1.0 branch, and then bumped the revno. See https://github.com/zmap/zmap/commit/c1f7959b90735f5bb4d473bd8958c8106785484b. Can you confirm that commit looks good on your end? If so, I'll tag it as 2.1.1 and we can pretend the packaging disaster that was 2.1.0 never happened. :grin:

cicku:
I will test this later. It should be fine, but give me some time. (4hrs probably since I'm outside)

cicku:
Test passed. :+1: for 2.1.1.

#287 should be included before release as well.

dadrian:
v2.1.1 is now tagged.

dadrian:
Thanks for all the help, if we're ever in the same city I owe you a beer. :beers:

cicku:
Sure, I will remember that :+1:

PR #288, opened 2015-09-11
Fix warning from [-Wfloat-equal]

cicku:
src/monitor.c: In function 'export_stats':
src/monitor.c:188:13: warning: comparing floating point with == or != is unsafe [-Wfloat-equal]
         if (!intrnl->min_hitrate_start) {
             ^
src/monitor.c:194:9: warning: comparing floating point with == or != is unsafe [-Wfloat-equal]
     if (intrnl->min_hitrate_start) {

---

Notes:

I'm not sure if this try is needed, or just use GCC #pragma instead. It's not a big issue, not sure if my way is proper or not.

### DIFF
--- a/src/monitor.c
+++ b/src/monitor.c
@@ -13,6 +13,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <unistd.h>
+#include <float.h>
 #include <math.h>
 #include <time.h>
 #include <sys/time.h>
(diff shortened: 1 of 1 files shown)


dadrian:
I believe @zakird has a fix for this in the DNS branch?

zakird:
That doesn't fix this exact issue, which oddly I wasn't seeing on my machine last night. So I'm going to merge it.

zakird:
I had to revert this a moment. Ago, this code prevents any scans from running. We'll need to figure out a solution. For now though, master is back to working.

cicku:
OK.

Or simply ignore this warning.

cicku:
I will take some time looking into this later, EPSILON might be the crux.

Honestly, I always ignore this warning...

dadrian:
I haven't taken a close look, but likely the right thing to do here is an epsilon comparison. We're just getting lucky with zero comparisons since I believe the variable is zero initialized and we only need this condition to fire at the start of a scan.

PR #292, opened 2015-09-15
Made output filters support fields with underscore or hyphen

### LINKED ISSUE #291: Output filter doesn't support fields with underscore or hyphen
Quite a lot of fields contains underscore/hyphens, e.g. `saddr-raw`, `icmp_responder`, and the field I especially wanted to use: `udp_pkt_size`.

But the token `T_FIELD`  in the filter grammar only matches alphanumeric characters, thus these fields cannot be used in output filters.

BobuSumisu:
Simply extended the lexer grammar to make `T_FIELD` also match underscore and hyphen (`[a-zA-Z][-_a-zA-Z0-9]+`).

Resolves #291.

### DIFF
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -31,6 +31,6 @@
 "<"                  return '<';
 "("                  return '(';
 ")"                  return ')';
-[a-zA-Z][a-zA-Z0-9]+ yylval.string_literal = strdup(yytext); return T_FIELD;
+[a-zA-Z][-_a-zA-Z0-9]+ yylval.string_literal = strdup(yytext); return T_FIELD;
 
 %%


dadrian:
I think @zakird may have simultaneously fixed this in a (currently unmerged) branch yesterday?

BobuSumisu:
Oh, didn't see that commit! @zakird maybe add a hyphen to the lexer rule as well to support fields with hyphens in them in output-filters?

zakird:
I happened to change this in branch the same day to support underscores (which is a crazy coincidence). However the change isn't in master yet, so I'm happy to merge this. I'll deal with the conflict later.

PR #295, opened 2015-09-21
DNS Probe Module (JSON-only)

dadrian:


src/lexer.l
```diff
@@ -31,6 +31,6 @@
 "<"                  return '<';
 "("                  return '(';
 ")"                  return ')';
-[a-zA-Z][a-zA-Z0-9]+ yylval.string_literal = strdup(yytext); return T_FIELD;
+[a-zA-Z][a-zA-Z0-9_]+ yylval.string_literal = strdup(yytext); return T_FIELD;
```
dadrian: :heart:

src/output_modules/module_json.c
```diff
@@ -50,7 +50,7 @@ int json_output_file_init(struct state_conf *conf, UNUSED char **fields,
 	return EXIT_SUCCESS;
 }
 
-char *hex_encode(char *packet, int buflen)
+char *hex_encode(unsigned char *packet, int buflen)
```
dadrian: If we're making this unsigned, we should make it a uint8_t

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    qtype_qtype_to_strid[DNS_QTYPE_MX] = 5;
+    qtype_qtype_to_strid[DNS_QTYPE_TXT] = 6;
+    qtype_qtype_to_strid[DNS_QTYPE_AAAA] = 7;
+    qtype_qtype_to_strid[DNS_QTYPE_RRSIG] = 8;
+    qtype_qtype_to_strid[DNS_QTYPE_ALL] = 9;
+}
+
+uint16_t _qtype_str_to_code(char* str) 
+{
+    for (int i = 0; i < (int) (sizeof(qtype_strs)/sizeof(qtype_strs[0])); i++) {
```
dadrian: Extract this to another global variable?

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+#include "probe_modules.h"
+#include "packet.h"
+#include "logger.h"
+#include "module_dns.h"
+#include "module_udp.h"
+#include "../fieldset.h"
+
+#define DNS_SEND_LEN 512    // This is arbitrary 
+#define UDP_HEADER_LEN 8
+#define PCAP_SNAPLEN 3000   // This is even more arbitrary
```
dadrian: This seems kind of large. 3K per incoming packet is a lot.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+
+    memcpy(qname_p, qname, qname_len);
+
+    tail_p->qtype = htons(qtype);
+    tail_p->qclass = htons(0x01); // MAGIC NUMBER. Let's be honest. This is only ever 1
+
+    return EXIT_SUCCESS;
+}
+
+uint16_t _get_name_helper(char* data, uint16_t data_len, char* payload,
```
zakird: This should be defined static.

src/lexer.l
zakird: This is going to conflict in master.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+#include <string.h>
+#include <assert.h>
+
+#include "../../lib/includes.h"
+#include "../../lib/random.h"
+#include "../../lib/xalloc.h"
+#include "probe_modules.h"
+#include "packet.h"
+#include "logger.h"
+#include "module_dns.h"
```
zakird: put include in rational location

src/probe_modules/module_dns.c
zakird: can we make this.... less arbitrary? 3K is gigantic.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+{
+    for (int i = 0; i < (int) (sizeof(qtype_strs)/sizeof(qtype_strs[0])); i++) {
+        if (strcmp(qtype_strs[i], str) == 0)
+            return qtype_strid_to_qtype[i];
+    }
+    return 0;
+}
+
+
+static uint16_t _domain_to_qname(char** qname_handle, char* domain)
```
dadrian: Can we domain be a `const char *`?

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+typedef uint8_t bool;
+
+// zmap boilerplate
+probe_module_t module_dns;
+static int num_ports;
+
+const char default_domain[] = "www.google.com";
+const uint16_t default_qtype = DNS_QTYPE_A; 
+
+static char *dns_packet = NULL;
```
zakird: don't we give you a buffer to do this in?

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+            }
+            qname[i] = j - i - 1;
+        }
+    }
+    *qname_handle = qname;
+    assert((*qname_handle)[len-1] == '\0');
+    return len;
+}
+
+static int _build_global_dns_packet(char* domain)
```
dadrian: const

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+        "RRSIG",
+        "ALL"
+};
+
+const dns_qtype qtype_strid_to_qtype[] = { DNS_QTYPE_A, DNS_QTYPE_NS, DNS_QTYPE_CNAME, 
+        DNS_QTYPE_SOA, DNS_QTYPE_PTR, DNS_QTYPE_MX, DNS_QTYPE_TXT, 
+        DNS_QTYPE_AAAA, DNS_QTYPE_RRSIG, DNS_QTYPE_ALL
+};
+
+#define DNS_QR_ANSWER   1
```
zakird: move this to be by the rest of the #defines.

src/probe_modules/module_dns.c
zakird: This is bad form. Define length of a variable.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    dns_packet_len = sizeof(dns_header) + qname_len + sizeof(dns_question_tail);
+
+    if (dns_packet_len > DNS_SEND_LEN) {
+            log_fatal("dns", "DNS packet bigger (%d) than our limit (%d)", 
+                    dns_packet_len, DNS_SEND_LEN);
+            return EXIT_FAILURE;
+    }
+    
+    dns_packet = xmalloc(dns_packet_len);
+    memset(dns_packet, 0x00, dns_packet_len);
```
zakird: redundant. This is done by xmalloc.

dadrian: This is redundant. `xmalloc` zeros memory.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+
+static uint16_t _domain_to_qname(char** qname_handle, char* domain) 
+{   
+    // String + 1byte header + null byte
+    uint16_t len = strlen(domain) + 1 + 1;
+    char* qname = xmalloc(len);
+    // Add a . before the domain. This will make the following simpler.
+    qname[0] = '.';
+    // Move the domain into the qname buffer.
+    memcpy(qname + 1, domain, strlen(domain));
```
zakird: Do you expect a nullbyte? You are not guaranteed this with a memcpy. Why not use strcpy?

src/probe_modules/module_dns.c
dadrian: Most of these parameters can probably be `const char *`

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    qtype_qtype_to_strid[DNS_QTYPE_SOA] = 3;
+    qtype_qtype_to_strid[DNS_QTYPE_PTR] = 4;
+    qtype_qtype_to_strid[DNS_QTYPE_MX] = 5;
+    qtype_qtype_to_strid[DNS_QTYPE_TXT] = 6;
+    qtype_qtype_to_strid[DNS_QTYPE_AAAA] = 7;
+    qtype_qtype_to_strid[DNS_QTYPE_RRSIG] = 8;
+    qtype_qtype_to_strid[DNS_QTYPE_ALL] = 9;
+}
+
+uint16_t _qtype_str_to_code(char* str)
```
zakird: This function should probably be static.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    dns_question_tail* tail_p = (dns_question_tail*)(dns_packet + sizeof(dns_header) + qname_len);
+
+    // All other header fields should be 0. Except id, which we set per thread.
+    dns_header_p->rd = 1; // Is one bit. Don't need htons
+    dns_header_p->qdcount = htons(1);
+
+    memcpy(qname_p, qname, qname_len);
+
+    tail_p->qtype = htons(qtype);
+    tail_p->qclass = htons(0x01); // MAGIC NUMBER. Let's be honest. This is only ever 1
```
zakird: can you comment what these variables are? I have no idea offhand.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+                return 0;
+            }
+
+            // We need to add a dot if we are:
+            // -- Not first level recursion.
+            // -- have consumed bytes
+            if (recursion_level > 0 || bytes_consumed > 0) {
+
+                if (name_len < 1) {
+                    log_warn("dns", "Exceeded static name field allocation.");
```
dadrian: Is there a big reason not to just make this dynamic? I guess performance?

zakird: I'd log an IP address here. This will help whoever hits this.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+                name++;
+                name_len--;
+            }
+
+            // Now we've consumed a byte.
+            ++bytes_consumed;
+
+            // Did we run out of our arbitrary buffer?
+            if (byte > name_len) {
+                log_warn("dns", "Exceeded static name field allocation.");
```
zakird: log IP address for further troubleshooting.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+            // -- have consumed bytes
+            if (recursion_level > 0 || bytes_consumed > 0) {
+
+                if (name_len < 1) {
+                    log_warn("dns", "Exceeded static name field allocation.");
+                    return 0;
+                }
+            
+                name[0] = '.';
+                name++;
```
dadrian: This will get compiled identically in this case, but I'm partial to doing preincrement here. ¯\_(ツ)_/¯

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    // -- a null byte, which terminates
+    // -- a segment length which either terminates or ensures we keep looping
+
+    assert(0);
+    return 0;
+}
+
+// data: Where we are in the dns payload
+// payload: the entire udp payload
+char* _get_name(char* data, uint16_t data_len, char* payload,
```
zakird: static please.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+
+// data: Where we are in the dns payload
+// payload: the entire udp payload
+char* _get_name(char* data, uint16_t data_len, char* payload, 
+        uint16_t payload_len, uint16_t* bytes_consumed)
+{
+    log_trace("dns", "call to _get_name, data_len: %d", data_len);
+
+    char* name = xmalloc(MAX_NAME_LENGTH);
+    memset(name, 0x00, MAX_NAME_LENGTH);
```
zakird: not needed

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    // data is handle to the start of this RR
+    // data_len is a pointer to the how much total data we have to work with.
+    // This is awful. I'm bad and should feel bad.
+    uint16_t bytes_consumed = 0;
+
+    char* question_name = _get_name(*data, *data_len, payload, payload_len,  &bytes_consumed);
+
+    // Error.
+    if (question_name == NULL) {
+        return 1;
```
zakird: use EXIT_FAILURE and EXIT_SUCCESS instead of 0 or 1

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+                name_len--;
+            }
+
+            uint16_t rec_bytes_consumed = _get_name_helper(payload + offset, 
+                    payload_len - offset, payload, payload_len, name, name_len, 
+                    recursion_level + 1);
+
+            // We are done so don't bother to increment the pointers.
+            if (rec_bytes_consumed == 0) {
+                log_trace("dns", "_get_name_helper OUT. rec level %d failed", recursion_level);
```
dadrian: There's a lot of log_trace's here, which might have a non-negligible performance impact since we don't actually optimize out the calls when trace-level logs aren't output.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    // Error.
+    if (question_name == NULL) {
+        return 1;
+    }
+
+    assert(bytes_consumed > 0);
+
+    if ( (bytes_consumed + sizeof(dns_question_tail)) > *data_len) {
+        free(question_name);
+        return 1;
```
zakird: use EXIT_FAILURE and EXIT_SUCCESS instead of 0 or 1

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    fs_add_uint64(qfs, "qclass", qclass);
+
+    // Now we're adding the new fs to the list.
+    fs_add_fieldset(list, "question", qfs);
+
+    // Now update the pointers.
+    *data = *data + bytes_consumed + sizeof(dns_question_tail);
+    *data_len = *data_len - bytes_consumed - sizeof(dns_question_tail);
+
+    return 0;
```
zakird: use EXIT_FAILURE and EXIT_SUCCESS instead of 0 or 1

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+}
+
+// XXX This should be merged with _process_response_question. 
+// Ended up being almost exactly the same
+bool _process_response_answer(char **data, uint16_t* data_len, char* payload,
```
zakird: static

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    // data is handle to the start of this RR
+    // data_len is a pointer to the how much total data we have to work with.
+    // This is awful. I'm bad and should feel bad.
+    uint16_t bytes_consumed = 0;
+
+    char* answer_name = _get_name(*data, *data_len, payload, payload_len,  &bytes_consumed);
+
+    // Error.
+    if (answer_name == NULL) {
+        return 1;
```
zakird: use EXIT_FAILURE and EXIT_SUCCESS instead of 0 or 1

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+
+    uint16_t type = ntohs(tail->type);
+    uint16_t class = ntohs(tail->class);
+    uint32_t ttl = ntohl(tail->ttl);
+    uint16_t rdlength = ntohs(tail->rdlength);
+    char* rdata = tail->rdata;
+
+    if ((rdlength + bytes_consumed + sizeof(dns_answer_tail)) > *data_len) {
+        free(answer_name);
+        return 1;
```
zakird: use EXIT_FAILURE and EXIT_SUCCESS instead of 0 or 1

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    // Error.
+    if (answer_name == NULL) {
+        return 1;
+    }
+
+    assert(bytes_consumed > 0);
+
+    if ( (bytes_consumed + sizeof(dns_answer_tail)) > *data_len) {
+        free(answer_name);
+        return 1;
```
zakird: use EXIT_FAILURE and EXIT_SUCCESS instead of 0 or 1

src/probe_modules/module_dns.h
```diff
@@ -0,0 +1,79 @@
...
+    DNS_QTYPE_SOA   = 6,
+    DNS_QTYPE_PTR   = 12,
+    DNS_QTYPE_MX    = 15,
+    DNS_QTYPE_TXT   = 16,
+    DNS_QTYPE_AAAA  = 28,
+    DNS_QTYPE_RRSIG = 46,
+    DNS_QTYPE_ALL   = 255
+} dns_qtype;
+
+// TODO: we may want to put out strings.
```
zakird: ??

src/probe_modules/module_dns.h
```diff
@@ -0,0 +1,79 @@
...
+
+typedef struct __attribute__((packed)) {
+    uint16_t type;
+    uint16_t class;
+    uint32_t ttl;
+    uint16_t rdlength;
+    char rdata[]; 
+} dns_answer_tail;
+
+// XXX
```
zakird: What's the problem here?

src/probe_modules/module_dns.c
dadrian: Redundant.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+
+        fs_add_uint64(fs, "dns_unconsumed_bytes", 0); 
+        fs_add_uint64(fs, "dns_parse_err", 1); 
+        fs_add_binary(fs, "raw_data", len, (char*)packet, 0);
+        
+        return;
+
+    } else {
+        // This should not happen. Both the pcap filter and validate packet prevent this.
+        assert(0);
```
zakird: please log_fatal instead of assert(0) so that a log message is available.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+        
+        // ICMP info
+        fs_add_null(fs, "icmp_responder");
+        fs_add_null(fs, "icmp_type");
+        fs_add_null(fs, "icmp_code");
+        fs_add_null(fs, "icmp_unreach_str");
+        
+        if (!is_valid) {
+            // DNS header
+            fs_add_null(fs, "dns_id");
```
zakird: because this is going to be JSON, putting dns_* in front of every field will actually lead to a negligible increase in file sizes. Like 2.5 GB worth for a scan that returns 30 million responses. This will end up resolving itself when the file is compressed, but we should consider whether they're actually necessary.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    
+    // Build our new question fieldset
+    fieldset_t *qfs = fs_new_fieldset(); 
+    fs_add_string(qfs, "name", question_name, 1);
+    fs_add_uint64(qfs, "qtype", qtype);
+    if (qtype > MAX_QTYPE || qtype_qtype_to_strid[qtype] == BAD_QTYPE_VAL) {
+        fs_add_string(qfs, "qtype_str", (char*) BAD_QTYPE_STR, 0);
+    } else {
+        // I've written worse things than this 3rd arg. But I want to be fast.
+        fs_add_string(qfs, "qtype_str", (char*)qtype_strs[qtype_qtype_to_strid[qtype]], 0);
```
dadrian: :disappointed:

src/probe_modules/module_dns.c
dadrian: How hard is this to merge with `_process_response_question`? We should really do this now. In all reality, we're never going to fix this later.

src/probe_modules/module_dns.c
dadrian: Should be `EXIT_FAILURE`

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    // Our memset ensured null byte.
+    assert(name[MAX_NAME_LENGTH - 1] == '\0');             
+
+    log_trace("dns", "return success from _get_name, bytes_consumed: %d, string: %s", *bytes_consumed, name);
+
+    return name;
+
+}
+
+static bool _process_response_question(char **data, uint16_t* data_len, char* payload,
```
zakird: 80 char please

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+/*
+ * Start of required zmap exports.
+ */
+
+static int dns_global_initialize(struct state_conf *conf) 
+{
+    char *qtype_str = NULL;
+    char *domain = NULL;
+
+    // This is zmap boilerplate. Why do I have to write this?
```
dadrian: @paul-pearce pull requests welcome :point_right:

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+    free(qtype_str);
+
+    return _build_global_dns_packet(domain);
+}
+ 
+int dns_global_cleanup(UNUSED struct state_conf *zconf, 
+        UNUSED struct state_send *zsend, UNUSED struct state_recv *zrecv) 
+{
+    if (dns_packet) {
+        free(dns_packet);
```
dadrian: Nice.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+            return EXIT_FAILURE;
+        }
+    }
+
+    free(qtype_str);
+
+    return _build_global_dns_packet(domain);
+}
+ 
+int dns_global_cleanup(UNUSED struct state_conf *zconf,
```
dadrian: This might be the first cleanup, ever.

zakird: woah. you can cleanup stuff in zmap? who knew.

src/output_modules/module_json.c
paul-pearce: The hex_encode functions in all other output modules follow this format (which is why I used it). I actually think forcing it unsigned is the wrong answer. You should be using %.2hhx which makes the signage irreverent.

src/probe_modules/module_dns.c
zakird: static

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+        uint32_t min_len = 4*ip_hdr->ip_hl + ICMP_UNREACH_HEADER_SIZE
+                + sizeof(struct ip) + sizeof(struct udphdr);
+        if (len < min_len) {
+            // Not enough information for us to validate
+            return 0;
+        }
+
+        struct icmp *icmp = (struct icmp*) ((char *) ip_hdr + 4*ip_hdr->ip_hl);
+        // We want to handle more of this.
+        /*if (icmp->icmp_type != ICMP_UNREACH) {
```
dadrian: Why is this commented out?

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+
+            // Did we parse OK?
+            fs_add_uint64(fs, "dns_parse_err", err); 
+        }
+    
+        // Now the raw stuff.
+        fs_add_binary(fs, "raw_data", (udp_len - sizeof(struct udphdr)), (void*) &udp_hdr[1], 0);
+        //fs_add_binary(fs, "raw_data", len, (void*)packet, 0);
+   
+        return;
```
dadrian: So other fields got added in other sections. I assume this is all part of the new recursive field goodness, but are we handling the case in a failed packet where one of the other DNS functions adds in a field (e.g. `"name"`)?.

src/output_modules/module_json.c
zakird: @dadrian why?

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+
+    } else {
+        // This should not happen. Both the pcap filter and validate packet prevent this.
+        assert(0);
+        return;
+    }
+}
+
+static fielddef_t fields[] = {
+    {.name = "classification", .type="string", .desc = "packet protocol"},
```
dadrian: Fun fact: this syntax is actually a GNU extension. Clang also supports it, so I'm all for using it.

src/probe_modules/module_dns.c
zakird: longterm we should probably #define something.

src/probe_modules/module_dns.h
```diff
@@ -0,0 +1,79 @@
...
+    unsigned cd :1;    /* checking disabled */
+    unsigned ad :1;    /* authenticated data */
+    unsigned z:1;      /* reserved set to 0 */
+    unsigned ra:1;     /* recursion available */
+
+    uint16_t qdcount;  /* # entries in question section */
+    uint16_t ancount;  /* # RR in answer section */
+    uint16_t nscount;  /* # name server RR in authority section */
+    uint16_t arcount;  /* # RR in additional information section */
+} dns_header;
```
dadrian: Types should really end in `_t`, but I guess we can let this pass.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+            payload_len, name, MAX_NAME_LENGTH - 1, 0);
+
+    if (*bytes_consumed == 0) {
+        free(name);
+        return NULL;
+    }
+
+    // Our memset ensured null byte.
+    assert(name[MAX_NAME_LENGTH - 1] == '\0');             
+
```
zakird: omg newlines

src/probe_modules/module_dns.h
```diff
@@ -0,0 +1,79 @@
...
+    uint16_t qtype;
+    uint16_t qclass;
+} dns_question_tail;
+
+typedef struct __attribute__((packed)) {
+    uint16_t type;
+    uint16_t class;
+    uint32_t ttl;
+    uint16_t rdlength;
+    char rdata[];
```
dadrian: Why not `uint8_t`? Also, is this a VLA or shouldn't this be a `char *`?

src/probe_modules/module_dns.c
paul-pearce: Agreed. Done.

src/probe_modules/module_dns.h
```diff
@@ -0,0 +1,79 @@
...
+    DNS_QTYPE_TXT   = 16,
+    DNS_QTYPE_AAAA  = 28,
+    DNS_QTYPE_RRSIG = 46,
+    DNS_QTYPE_ALL   = 255
+} dns_qtype;
+
+// TODO: we may want to put out strings.
+typedef enum {
+	DNS_RCODE_NOERR        = 0,   /* Response code NO ERROR */
+	DNS_RCODE_FORMATERR    = 1,   /* Response code NO ERROR */
```
dadrian: `/* Response code NO ERROR */` these are not all NO ERROR

src/probe_modules/module_dns.c
paul-pearce: After off-list discussion moving to 1500. (Note, rrsig's for google, for example, are upwards of 1000 bytes.)

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1043 @@
...
+}
+
+int dns_validate_packet(const struct ip *ip_hdr, uint32_t len,
+        uint32_t *src_ip, uint32_t *validation)
+{
+    // This does the heavy lifting.
+    if (!udp_validate_packet(ip_hdr, len, src_ip, validation)) {
+        return 0;
+    }
+
```
zakird: Is there a reason not to check (1) if this is big enough to be a DNS packet and (2) the id field?

src/output_modules/module_json.c
dadrian: @zakird I come from the "bytes are unsigned 8-bit integers, not characters" camp.

@paul-pearce You're probably right about `%.2hhx`, but at the same time this is exactly why `printf` has been known to induce tears.

src/probe_modules/module_dns.c
dadrian: Agreed. This isn't Python. We should remove the leading underscores from the "private" functions, and declare them all as static. Internal linkage FTW.

src/probe_modules/probe_modules.c
```diff
@@ -26,15 +26,17 @@ extern probe_module_t module_icmp_echo_time;
...
 extern probe_module_t module_upnp;
+extern probe_module_t module_dns;
 // ADD YOUR MODULE HERE
 
 probe_module_t* probe_modules[] = {
-	&module_tcp_synscan,
-	&module_icmp_echo,
-	&module_icmp_echo_time,
-	&module_udp,
+    &module_tcp_synscan,
```
zakird: Did you take our tabs and make them spaces? :-1:

src/probe_modules/module_dns.c
dadrian: agreed

src/probe_modules/module_dns.c
dadrian: :+1:

src/probe_modules/module_dns.c
paul-pearce: done.

src/probe_modules/module_dns.c
dadrian: I'm not sure if this can be static since it's exposed in that struct.

src/probe_modules/module_dns.c
dadrian: Given that zmap has metadata output and can only scan one protocol per scan, I don't see why we would ever prefix a fieldname that isn't duplicated between OSI layers.

src/probe_modules/probe_modules.c
dadrian: :no_good:

src/lexer.l
paul-pearce: It did. I resolved the conflict keeping the master version.

src/probe_modules/module_dns.c
paul-pearce: I think I snagged all these.

src/probe_modules/module_dns.c
paul-pearce: Done

src/probe_modules/module_dns.c
paul-pearce: Done

src/probe_modules/module_dns.c
paul-pearce: done

src/probe_modules/module_dns.c
paul-pearce: Removed throughout.

src/probe_modules/module_dns.c
paul-pearce: Static done.

Renaming begrudgingly done.

src/probe_modules/module_dns.c
paul-pearce: No. (Discussed over chat. Zakir begrudgingly agrees.)

src/probe_modules/module_dns.c
paul-pearce: Done.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1039 @@
...
+                    payload_len - offset, payload, payload_len, name, name_len, 
+                    recursion_level + 1);
+
+            // We are done so don't bother to increment the pointers.
+            if (rec_bytes_consumed == 0) {
+                log_trace("dns", "_get_name_helper OUT. rec level %d failed", recursion_level);
+                return 0;
+            } else {
+                bytes_consumed += 2;
+                log_trace("dns", "_get_name_helper OUT. rec level %d success. %d rec bytes consumed. %d bytes consumed.", recursion_level, rec_bytes_consumed, bytes_consumed);
```
dadrian: Line break after the string literal please.

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1039 @@
...
+static bool process_response_question(char **data, uint16_t* data_len, const char* payload,
+        uint16_t payload_len, fieldset_t* list)
+{   
+    // Payload is the start of the DNS packet, including header
+    // data is handle to the start of this RR
+    // data_len is a pointer to the how much total data we have to work with.
+    // This is awful. I'm bad and should feel bad.
+    uint16_t bytes_consumed = 0;
+
+    char* question_name = get_name(*data, *data_len, payload, payload_len,  &bytes_consumed);
```
dadrian: This is over 80 characters. Can fix the 80 character limit for all lines that are not string literals?

src/probe_modules/module_dns.c
paul-pearce: Done.

zakird:
OK Great. Now solve the following and we're set:

kccpkbgufdphqtydfinrrtmvgrkdnbvfdetdgiltxrgud dkotsmbpvgegltgckqracqcwdnawcrxizakftlewrptyc qkyvxchastponcqqrhjvajuwetmcmspkqdyhjvdahctrl svskcgfzqddzxgsfrlswcgsjtbhafsiasprjahkjrjumv gkmitzhfpdispzlvlgwtfplkkevbngcebshctjrwxbafs pezqnrwxcvyggaonwddkackawbbikftiovkcgghjvlnhi ffsqesvycklvrwbbirepbbvfexoscdygzwpfdtkfqiy cwhjvlnhiqibskhjvnpist

src/probe_modules/module_dns.c
paul-pearce: @dadrian Performance was why. So I'd like to leave this as is.

@zakird IP is not available here. Discussed off list. Agreed to not log.

src/probe_modules/module_dns.c
paul-pearce: Closed. Won't fix (discussed offlist)

src/probe_modules/module_dns.c
paul-pearce: Resolved

src/probe_modules/module_dns.c
```diff
@@ -0,0 +1,1039 @@
...
+        // App success: has qr and rcode bits right
+        // Any app level parsing issues: dns_parse_err
+
+        // High level info
+        fs_add_string(fs, "classification", (char*) "dns", 0);
+        fs_add_uint64(fs, "success", is_valid);
+        fs_add_uint64(fs, "app_success", is_valid && (qr == DNS_QR_ANSWER) && (rcode == DNS_RCODE_NOERR));
+        
+        // UDP info
+        fs_add_uint64(fs, "udp_sport", ntohs(udp_hdr->uh_sport));
```
zakird: This should be sport and dport to match TCP.

src/probe_modules/module_dns.c
paul-pearce: Done

src/probe_modules/module_dns.c
paul-pearce: Removed.

src/probe_modules/module_dns.c
paul-pearce: Fixed static.

Honestly. The two are different enough now that I'm fine with them being separate unless there's huge pushback.

src/probe_modules/module_dns.c
paul-pearce: Resolved.

src/probe_modules/module_dns.c
paul-pearce: Resolved.

src/probe_modules/module_dns.c
paul-pearce: Resolved.

src/probe_modules/module_dns.c
paul-pearce: Resolved.

src/probe_modules/module_dns.h
paul-pearce: Great question. Removed. I think that was the first thing I wrote in this entire mess. (out->our)

src/probe_modules/module_dns.h
paul-pearce: It should be removed. Done.

src/probe_modules/module_dns.c
paul-pearce: done

src/probe_modules/module_dns.c
paul-pearce: Resolved.

src/probe_modules/module_dns.c
paul-pearce: I basically bit a mini hash table. Don't judge me. It's the right answer here.

src/probe_modules/module_dns.c
paul-pearce: On my todo. :)

src/probe_modules/module_dns.c
paul-pearce: I try.

src/probe_modules/module_dns.c
paul-pearce: It can be and is (without issue).

Done.

src/probe_modules/module_dns.c
paul-pearce: I was experimenting with handling other kinds of ICMP. The experiment worked and I forgot to remove the commented code (forgot to XXX it).

done.

src/probe_modules/module_dns.c
paul-pearce: Yes that is handled. The way it's handled is the top level schema is just a list of things. Then each thing has a schema of an RR. There's no code path that creates differing fields (or if there is it's a bug we didn't hit yet).

src/probe_modules/module_dns.c
paul-pearce: (nothing to do)

src/probe_modules/module_dns.h
paul-pearce: resolved.

src/probe_modules/module_dns.c
paul-pearce: Dear @zakird 

I find that newlines really help me understand code when what it's doing is ungodly awful (e.g. dns).

What do you think?

Thanks.

src/probe_modules/module_dns.h
paul-pearce: It is indeed a VLA. I can be talked into making a uint8_t but that seemed not to be the project standard.

src/probe_modules/module_dns.h
paul-pearce: Legacy bug. Done.

src/probe_modules/module_dns.c
paul-pearce: We check (1) below. We save (2) for process packet (as we discussed once upon a time).

src/probe_modules/probe_modules.c
paul-pearce: Bug. Will be fixed in the retab commit.

src/probe_modules/module_dns.c
paul-pearce: Resolved.

src/probe_modules/module_dns.c
paul-pearce: Yep. Will resolve.

src/probe_modules/module_dns.c
paul-pearce: Done.

Issue #301, opened 2015-10-25
ZMap does not guarantee UTF-8 


zakird:
ZMap currently outputs non-UTF8 encoded data. This hasn't been a problem previously because we constructed most of the strings. However, this model broken when the DNS module was introduced. We need to ensure that strings are UTF-8 valid (by escaping any invalid characters). A couple of notes:

1) json-c does not escape invalid UTF-8 (so therefore doesn't save us). As well, we really want every output module to be outputting real strings

2) This is what Golang does. We should probably do something similar "String values encode as JSON strings coerced to valid UTF-8, replacing invalid bytes with the Unicode replacement rune. The angle brackets "<" and ">" are escaped to "\u003c" and "\u003e" to keep some browsers from misinterpreting JSON output as HTML. Ampersand "&" is also escaped to "\u0026" for the same reason."

3) I'm attaching a file which is known to cause problems and can be used for testing. We can test by trying to parse the JSON with python.

[out.txt](https://github.com/zmap/zmap/files/20331/out.txt)

zakird:
Paul I'm assigning this to you for now, but anyone can grab it.

paul-pearce:
I changed the milestone to 3.0 given that this needs to pull in a new library.

zakird:
We should use standard behavior for invalid characters:


pearce [3:37 PM] 
>>>s = chr(0x6a) + chr(0xc0) + chr(0x0c)
>>> s
'j\xc0\x0c'
>>> s.decode("utf-8", "replace")
u'j\ufffd\x0c'

pearce [3:37 PM]
it replaces it with the rune.

zakir [3:38 PM]

paul-pearce:
Expanding on Zakir's comment a bit. We had some discussion about the escaping behavior for the fix as deployed. We decided to change it and go with a "lossy" solution where the bad bytes are replaced with a �. This matches the behavior of most languages. Sadly. 

I will queue this fix up and change the code when I get a chance.

PR #302, opened 2015-10-26
Handle Unsafe utf8 strings

paul-pearce:
This addresses https://github.com/zmap/zmap/issues/301 by requiring that potentially unsafe strings be handled via the new fieldset operation fs_add_unsafe_string().

fs_add_unsafe_string() is designed to be used when you are supposed be given utf8 but the string may be malformed (e.g. you yanked it out of a DNS packet). fs_add_unsafe_string() will check if the string is a valid utf8 encoding. If not it will escape the offending byte as \u00XX.

A drawback here is we don't escape the escape character. This makes it impossible to differentiate between an escaped sequence and those bytes actually being in the string. Given that this is an extreme corner case (currently 1 server on the Internet, one protocol) I believe not altering the escape character is the right call. Especially given that any string emitted by this operation could just as easily be marked as a complete error and discarded.

Marked as ZMap 3.0 given that this pulls in a new library.

Note: TravisCI fails because the required library is not on the CI server.

### DIFF
--- a/src/fieldset.c
+++ b/src/fieldset.c
@@ -12,6 +12,7 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <assert.h>
+#include <unistr.h>
 
 #include "../lib/logger.h"
 #include "../lib/xalloc.h"
@@ -29,7 +30,7 @@ void gen_fielddef_set(fielddefset_t *fds, fielddef_t fs[], int len)
 fieldset_t *fs_new_fieldset(void)
 {
 	fieldset_t *f = xcalloc(1, sizeof(fieldset_t));
-    f->len = 0;
+	f->len = 0;
 	f->type = FS_FIELDSET;
 	return f;
 }
@@ -37,7 +38,7 @@ fieldset_t *fs_new_fieldset(void)
 fieldset_t *fs_new_repeated_field(int type, int free_)
 {
 	fieldset_t *f = xcalloc(1, sizeof(fieldset_t));
-    f->len = 0;
+	f->len = 0;
 	f->type = FS_REPEATED;
 	f->inner_type = type;
 	f->free_ = free_;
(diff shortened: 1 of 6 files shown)


src/fieldset.c
```diff
@@ -102,6 +103,64 @@ static void fs_modify_word(fieldset_t *fs, const char *name, int type,
...
+
+		assert(ptr >= buf);
+		assert(ptr < buf + strlen(buf));
+
+		ptr++;
+		i++;
+	}
+
+	// i is the total number of errors. We need 5 characters for each error \u00XX
+	char *safe_buf = xmalloc(strlen(buf) + i*5 + 1);
```
dadrian: \u00XX is 6 characters

src/fieldset.c
```diff
@@ -102,6 +103,64 @@ static void fs_modify_word(fieldset_t *fs, const char *name, int type,
...
 }
 
+static char *sanitize_utf8(const char *buf)
+{
+	const char *ptr = buf;
+
+	// Count how many errors we encounter
+	uint32_t i = 0;
+	// Upper bounds to ensure termination even if u8_check is unsafe
+	while (i < strlen(buf) && ptr < buf + strlen(buf)) {
```
dadrian: This looks like a `for` loop in disguise

dadrian: Maybe not.

dadrian:
Have you actually tested this on any invalid UTF-8?

src/fieldset.c
paul-pearce: That should actually say 5 additional. Correcting.

src/fieldset.c
paul-pearce: Your reaction is exactly why I wrote it as such. In really it really wants to be a while(ptr!=null), but with a maximum iteration safety belt in case u8_check behaves out of spec. I can rewrite it like a for if you'd like, but I think it makes more sense semantically as it is. Want me to change it?

paul-pearce:
Thanks for fixing travis. 

Yes I have tested this on invalid UTF-8. There is a single DNS server on the internet behaving out of spec and I've tested this branch against that server. I've not performed a comprehensive scan yet, since I lack the resources to do that.

PR #304, opened 2015-10-28
Replace bad utf8 chars with the utf8 rune instead of escaping

paul-pearce:


paul-pearce:
This addresses https://github.com/zmap/zmap/issues/301

PR #311, opened 2015-11-20
Add BACNet support

dadrian:
Do not merge yet, just creating this for tracking purposes.


dadrian:
@zakird Can you double check this? I think this is all I'm going to do within ZMap. More can be done in ZGrab. This is not something I want to do any more in C.

dadrian:
Looks good to me. :ship:

PR #312, opened 2015-12-11
Update test_harness.c

jakeczyz:
Added conditional compilation when json enabled. The build was failing when building -DWITH_JSON=OFF.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
 project(ZMAP C)
-set(ZMAP_VERSION 2.1.1)
+set(ZMAP_VERSION 2.1.1b)
 
 option(WITH_REDIS "Build with support for Redis DB" OFF)
 option(WITH_MONGO "Build with support for MongoDB" OFF)
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -48,7 +48,7 @@ In addition, you can get following packages to get further functionalities:
 
 * On RHEL- and Fedora-based systems by running:
    ```sh
-   sudo yum install cmake gmp-devel gengetopt libpcap-devel flex byacc json-c-devel libunistring
+   sudo yum install cmake gmp-devel gengetopt libpcap-devel flex byacc json-c-devel libunistring-devel
    ```
 
 * On Mac OS systems [Homebrew](http://brew.sh/):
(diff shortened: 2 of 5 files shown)


zakird:
I'm not quite sure this is the right way of fixing this. The tests don't really make any sense without JSON support here. I think that likely the better solution is to not build test_harness.c by default and to add a CMAKE flag that determines whether tests should be built or not. @dadrian what are your thoughts?

dadrian:
I think we should get rid of -DWITH_JSON, and just require the newer version of json-c for all builds. It's in apt on 14.04.

zakird:
I agree with this too. There's major ZMap functionality that requires JSON at this point (e.g., metadata output).

jakeczyz:
:( we run zmap on some low power/storage systems (that aren't
ubuntu/debian) and make no use of json output. So, this will mean more
hassle for us. Would love to keep this optional if it doesn't overly
complicate your life... But, understand if it means nore work for you...

Cheers,
Jake



On Saturday, December 12, 2015, Zakir Durumeric <notifications@github.com>
wrote:

> I agree with this too. There's major ZMap functionality that requires JSON
> at this point (e.g., metadata output).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zmap/zmap/pull/312#issuecomment-164190847>.
>


-- 
Jakub (Jake) Czyz
Doctoral Candidate
Network and Security Research Group
University of Michigan - EECS / CSE
http://nsrg.eecs.umich.edu
http://jczyz.com

zakird:
Jake, What OS are you using that doesn't have json-c packages?

jakeczyz:
It's available on the OSes we use (Fedora 21, and CentOS 7.1).  Somehow, I
thought it wasn't for some reason and was installing from source.  ID10T
error. Please ignore. Sorry.



On Sun, Dec 13, 2015 at 12:15 PM, Zakir Durumeric <notifications@github.com>
wrote:

> Jake, What OS are you using that doesn't have json-c packages?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zmap/zmap/pull/312#issuecomment-164278129>.
>



-- 
Jakub (Jake) Czyz
Doctoral Candidate
Network and Security Research Group
University of Michigan - EECS / CSE
http://nsrg.eecs.umich.edu
http://jczyz.com

dadrian:
@jakeczyz are we good closing this pull request then?

jakeczyz:
Sure, I personally don't need to install with -DWITH_JSON=OFF.  Just be aware that the build fails during test phase when that flag is passed to cmake, because of the test that is still attempted without this patch.   But, it doesn't affect me, so do as you will :-D. Thx!

PR #313, opened 2015-12-11
Add DragonFly BSD support

Rolinh:
Hi,

This patch adds DragonFly BSD support.


zakird:
Looks good offhand (although I don't have a way of immediately testing DragonFly). Thanks for the addition.

PR #315, opened 2015-12-29
log_warn code to help identify an uncommon utf8 sanitization problem

paul-pearce:


src/fieldset.c
```diff
@@ -134,7 +134,16 @@ static char *sanitize_utf8(const char *buf)
 		safe_ptr = (char*)u8_check((uint8_t*)safe_buf, strlen(safe_buf));
 
 		// This implies we had less errors than we should.
-		assert(safe_ptr != NULL);
+		// This is temporary debug code.
+		if (safe_ptr == NULL) {
+			log_warn("fieldset", "UTF8 Sanitization issue. %u errors, fell through iter %u. Orig: %s new: %s",
+				i, j, buf, safe_buf);
```
zakird: can you please indent properly

zakird:
I don't follow the i=j, but if it's right, I don't have major concerns.

src/fieldset.c
dadrian: fwiw, I'm fine with string literals going over the limit because it makes them easier to grep for.

paul-pearce: I specifically violated the 80char limit on the string literal because it was done elsewhere in the code on log msgs. The missing tab was a mistake when I retabbed. It's been corrected.

paul-pearce:
i=j is needed to keep the assertions outside the loop from failing. In this odd error case we've countered errors wrong. If we don't reset i our expected string length will be wrong.

PR #322, opened 2016-01-17
race condition in monitor thread opening file

zakird:


paul-pearce:
lgtm

dadrian:
So to clarify, this fixes the race condition between dropping privileges and opening the monitor output file by moving the `fopen` to the main thread, before `drop_privs` is called. Assuming that's correct, looks good to me.

PR #324, opened 2016-01-17
Update missing dependencies on Fedora based systems

Pablohn26:
fatal error: unistr.h: No such file or directory when trying to compile without libunistring-devel installed.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -48,7 +48,7 @@ In addition, you can get following packages to get further functionalities:
 
 * On RHEL- and Fedora-based systems by running:
    ```sh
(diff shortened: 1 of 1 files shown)


paul-pearce:
Are you sure libunistring is needed as well as libunistring-devel?

Pablohn26:
libunistring was added before, not sure if it is an actual dependency or a typo. But libunistring-devel is needed for sure :)
I will try to find out that

Pablohn26:
It does not matter if you put it or not because it is a dependency as well. 

selection_382

So yes, it is needed

dadrian:
@Pablohn26 yes, but we don't enumerate the subdependencies of any other package. To build ZMap, we need the `devel` package, so let's just list the `devel` package.

paul-pearce:
lgtm

Issue #329, opened 2016-02-02
UPnP module should use add_unsafe_string


dadrian:

zakird:
This was fixed in 6aff5b8 and landed in master a couple of weeks ago.

PR #330, opened 2016-02-02
use safe strings in upnp

zakird:


src/fieldset.c
```diff
@@ -206,6 +206,16 @@ void fs_chkadd_string(fieldset_t *fs, const char *name, char *value, int free_)
 	}
 }
 
+void fs_chkadd_unsafe_string(fieldset_t *fs, const char *name, char *value, int free_)
+{
+	if (value) {
+		fs_add_unsafe_string(fs, name, value, free_);
```
dadrian: This will currently add empty strings. Not sure if that's intended behavior or not.

zakird: Empty strings are empty strings and should be treated like empty strings.

dadrian:
Looks fine, aside from the one comment about empty strings.

src/fieldset.c
dadrian: To clarify: I mean empty strings do not result in an `fs_add_null`, which I'm not sure if is intended behavior or not.

zakird: right, which is what's semantically correct.

dadrian: Great. :shipit:

dadrian:
Assuming it builds, let's merge.

paul-pearce:
This is for issue https://github.com/zmap/zmap/issues/329

PR #332, opened 2016-03-02
Custom flags should be put in the front as preferred

Fale:


zakird:
This seems fine to me---I was not aware of the custom. @cicku, any concerns from your side?

Fale:
I guess @cicku is ok, since he [added a command](http://pkgs.fedoraproject.org/cgit/rpms/zmap.git/commit/?id=457aea64739d169aa7751ceb2683373fe56d3bf9) (the sed one) to do exactly this.

Also I'm not sure he will be able to reply soon since the Fedora Project have tried to get in touch with him for the last month with no success.

zakird:
touché. merging.

cicku:
Thx.

Issue #339, opened 2016-04-22
Some ip addresses are missed by ZMAP


BingoZ:
Hi, I find some addresses can be missed by ZMAP， like `125.39.240.113`, if you telnet its port 80, you can find it is open, but zmap can not scan it out in SYN mode. 
I capture the packet and guess it is because the syn packet sent out by ZMAP doesn`t set MSS in TCP option. Is this a little bug or just under your considerations?

dragonkid:
I think this may be the reason why SYN was dropped.
[https://live.paloaltonetworks.com/t5/Management-Articles/SYN-Packets-Without-MSS-Are-Dropped-With-SYN-Cookie-Enabled/ta-p/52351](url)

Zmap should send normal SYN packet to avoid this.

dadrian:
Thanks for the report, we'll look into getting this fixed.

zakird:
I suspect the right thing to do here is to do A/B testing with this option set and unset. If there are no hosts lost by setting it, then it's an easy change. If you do lose hosts, then we're going to have to figure out a better. @dragonkid any chance you've tried testing with both?

dragonkid:
I found the ip `202.108.152.150` which can be used for test.

I have tried to set the option, no hosts was lost.

Should I make a pull request?

zakird:
That would be perfect, thank you.

dragonkid:
@zakird  I am already make the pull request. Was that ok?

PR #342, opened 2016-05-03
fix issue #339

dragonkid:
A a new method `make_tcp_syn_header` was added to avoid bring new bugs.

This is my first pull request, so please be patient:)


zakird:
I did some testing tonight trying to decide whether or not we should merge this. In the end, this change doesn't look to have has no meaningful impact on the results of a ZMap TCP scan. 

For 10 trials targeting 1mil IPs on TCP/80, the new style probe found on average less hosts and student's T-Test resulted in t=0.883,  p=0.4 (effectively indicating no statistically significant change). Given that the change doesn't find new hosts, but increases TCP packet size by 7%, increasing both the time and bandwidth of a scan, I'm going to close this PR. 

If you want to submit this as a new probe module, I'd be fine merging that. I just don't think it warrants changing the original TCP probe module.

PR #349, opened 2016-06-08
Add CSV output support for repeated/fieldset fields

hkennedy:
I was writing a probe module I intended to use with the CSV output module but ran into trouble when I needed to print repeated fields.

I just added a (recursive) function in case there are multiple repeat levels (as well as to avoid adding the space at the end of cs_process() for every Repeated field/Field set).


dadrian:
We specifically don't support CSV for repeated fields, since the number of output fields wouldn't be constant. How are you handling this?

dadrian:
Since I haven't heard from you, I'm going to close this pull request. Feel free to reopen if you have an answer to my previous question. Thanks!

PR #352, opened 2016-06-23
Fix out-of-source builds

ogasser:
Use correct CMAKE dir variable to fix out-of-source builds.
Verified that in-source builds are still working.


dadrian:
Thanks!

PR #372, opened 2016-08-25
Accept additional synack RST responses, correct syn scan RST behavior

paul-pearce:
Accepts additional packets as valid as part of the SYNACK scan module. This involves setting additional validation bits in the ACK Field (which we should have been doing to begin with) and checking them.

Also part of the PR is correcting how zmap validates RST packets for SYN scans. Per the RFC, in SYN-SENT you should accept SYN+0 as well as SYN+1 packets. We previously only accepted SYN+1

Rolled into this is fixing a few lines whitespace where spaces were used instead of tabs.

### DIFF
--- a/src/probe_modules/module_tcp_synackscan.c
+++ b/src/probe_modules/module_tcp_synackscan.c
@@ -94,34 +94,26 @@ static int synackscan_validate_packet(const struct ip *ip_hdr, uint32_t len,
 		return 0;
 	}
 
-	// Note: The traditional flow of returning 0 on wrong and then returning 1
-	// at the end made this code an unreadable mess. So we're going a bit
-	// against the grain
-
-	if (htonl(tcp->th_ack) == htonl(validation[0]) + 1) {
-		// ACK = SEQ + 1 is OK
-		return 1;
-	} else if (tcp->th_flags & TH_RST) {
-		// If we have a reset flag, we can have a few cases.
-		if (htonl(tcp->th_seq) == htonl(validation[2]) + 0) {
-			// If SEQ = ACK RST is valid
-			return 1;
-		} else if (htonl(tcp->th_seq) == htonl(validation[2]) + 1) {
-			// If SEQ = ACK + 1 RST is also valid
-			return 1;
-		} else {
-			// All other cases the RST is invalid.
+	// We handle RST packets different than all other packets
+	if (tcp->th_flags & TH_RST) {
+		// A RST packet must have either:
+		//	1) resp(ack) == sent(seq) + 1, or
+		//	2) resp(seq) == sent(ack), or
+		//	3) resp(seq) == sent(ack) + 1
+		// All other cases are a failure.
+		if (htonl(tcp->th_ack) != htonl(validation[0]) + 1
+				&& htonl(tcp->th_seq) != htonl(validation[2])
+				&& htonl(tcp->th_seq) != (htonl(validation[2]) + 1)) {
 			return 0;
 		}
 	} else {
-		// All other options are invalid.
-		return 0;
+		// For non RST packets, we must have resp(ack) == sent(seq) + 1
+		if (htonl(tcp->th_ack) != htonl(validation[0]) + 1) {
+			return 0;
+		}
 	}
 
-	// DO NOT ADD ANY CHECKS HERE.
-	assert(NULL);
-
-	return 0;
+	return 1;
 }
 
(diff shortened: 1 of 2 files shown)


zakird:
👍

src/probe_modules/module_tcp_synackscan.c
```diff
@@ -38,9 +45,88 @@ static int synackscan_init_perthread(void* buf, macaddr_t *src,
...
+	// validate destination port
+	if (!check_dst_port(ntohs(dport), num_ports, validation)) {
+		return 0;
+	}
+
+	// Note: The traditional flow of returning 0 on wrong and then returning 1
+	// at the end made this code an unreadable mess. So we're going a bit
+	// against the grain
+
+	if (htonl(tcp->th_ack) == htonl(validation[0]) + 1) {
```
dadrian: Shouldn't we be checking flags here?

src/probe_modules/module_tcp_synackscan.c
```diff
@@ -38,9 +45,88 @@ static int synackscan_init_perthread(void* buf, macaddr_t *src,
...
+		// ACK = SEQ + 1 is OK
+		return 1;
+	} else if (tcp->th_flags & TH_RST) {
+		// If we have a reset flag, we can have a few cases.
+		if (htonl(tcp->th_seq) == htonl(validation[2]) + 0) {
+			// If SEQ = ACK RST is valid
+			return 1;
+		} else if (htonl(tcp->th_seq) == htonl(validation[2]) + 1) {
```
dadrian: Why not just make this an `||`?

src/probe_modules/module_tcp_synscan.c
```diff
@@ -107,14 +107,22 @@ int synscan_validate_packet(const struct ip *ip_hdr, uint32_t len,
 	}
 	// validate tcp acknowledgement number
 	if (htonl(tcp->th_ack) != htonl(validation[0])+1) {
-		return 0;
+
+		// RSTs do not need to +1 the ack
```
zakird: I think the way this logic is arranged is currently confusing. It would be clearer if we had something like
if RST: apply these rules
else: apply these other rules.

Right now, this is effectively,
apply rules, if this one random rule fails, but you're a RST, just ignore that branch of the logic. While these are arguably equivalent, I believe that the former is cleaner than the latter.

src/probe_modules/module_tcp_synackscan.c
dadrian: By which I mean, it's weird to me that the second part of this if statement is specifically packets with `TH_RST` set. What type of packets does this apply to? If it applies to all packets, does this even need to be an `else if` underneath?

src/probe_modules/module_tcp_synscan.c
dadrian: I agree, I don't follow this code at all. What are the "acceptable" cases here?

src/probe_modules/module_tcp_synackscan.c
```diff
@@ -38,9 +45,88 @@ static int synackscan_init_perthread(void* buf, macaddr_t *src,
...
+			// All other cases the RST is invalid.
+			return 0;
+		}
+	} else {
+		// All other options are invalid.
+		return 0;
+	}
+
+	// DO NOT ADD ANY CHECKS HERE.
+	assert(NULL);
```
dadrian: In general, I'm not sure what the "success" conditions are supposed to be here. What are they?

src/probe_modules/module_tcp_synackscan.c
```diff
@@ -38,9 +45,88 @@ static int synackscan_init_perthread(void* buf, macaddr_t *src,
...
+		} else if (htonl(tcp->th_seq) == htonl(validation[2]) + 1) {
+			// If SEQ = ACK + 1 RST is also valid
+			return 1;
+		} else {
+			// All other cases the RST is invalid.
+			return 0;
+		}
+	} else {
+		// All other options are invalid.
+		return 0;
```
dadrian: This doesn't need to be an `else`. You can just `return 0` and then not have to worry about the `assert(NULL)`.

zakird:
@dadrian and I are fine with what you're trying to accomplish this in PR. However, the logic is a bit convoluted in the current state (see comments). Once those issues have been addressed, this should be good to go.

zakird:
This looks fine to me. Tests are still running though, so waiting to merge.

PR #380, opened 2016-09-15
Add clang-format check to TravisCI build

dadrian:
I'm not totally sure we want this, but I got nerd sniped hard core on it yesterday.

This adds a `.clang-format` file to ZMap, written for Clang Formatter 4.0, and adds the commands `make format` and `make format-check`, which format, and check to see if formatting is required, respectively. `make format-check` is added to the TravisCI build process on OS X only, since current versions of Ubuntu do not have clang-format 4.0 available (Trusty have 3.5, Xenial has 3.8).

The formatting rules follow kernel style with one deviation: use 4-space tabs, spaces instead of tabs, rather than 8-space tab characters as tabs.

This also runs the formatter on the code. The non-formatting changes are in the following files:
- `.clang-format`
- `.travis.yml`
- `scripts/check_style.sh`
- `scripts/style.py`

@zakird @paul-pearce @ppannuto


zakird:
I don't really see much reason to deviate from the Kernel guide here. Sanely configured editors take care of spaces versus tabs and there's no benefit from switching, particularly when it would cause deviation from a known standard.

ppannuto:
I have about zero stake in this, but FWIW I agree with Zakir. 

Don't see a good readon to leave kernel style if the project is already there (or even at all really, it's what I expect C code to look like)

Do like Travis-enforced formatting of whatever you do though.

ppannuto:
Bro, do you even rebase?

Or a y'all a squash-on-merge shop?

ppannuto:
(In add seriousness, having changes touching all those lines in your history is probably not great)

dadrian:
It appears the the formatter is currently aligning function arguments when a function declaration spans more than one line, rather than just indenting lines 2-N twice. It's unclear if that is consistent with kernel style or not, it's certainly not what we were doing previously. Thoughts?

ppannuto:
Do you mean

```c
int main() {
	many_arg_func(int arg1
			int twotab,
			int version
			) {

	many_arg_func(int arg1
	              int aligned,
	              int version
	             ) {
}
```

That didn't copy/paste into gh markdown as well as hoped, to be clear that's:

```
<tab><tab><tab>int twotab
<tab><14 spaces>int aligned
```

Where `len('many_arg_func(') == 14`.

Not sure what kernel style is here, but IMHO the latter is the best b/c it works correctly no matter what you set the tabstop to.

dadrian:
I was referring to

```
// What we were previously doing:
//
// <function><args>
// <tab><tab><rest of args>
int man_arg_function(int argument_one, int argument_two, int argument_three,
                int argument_four);
```
versus
```
// What .clang-format seems to be doing currently, but could be changed.
// This may/may not actually be consistent with kernel style, it's unclear.
//
// <function><args>
// <tab><tab><spaces to align><rest of args>
int man_arg_function(int argument_one, int argument_two, int argument_three,
                     int argument_four);
```

I think I like our original approach better.

ppannuto:
Hrmph, the clang-format seems like the worst, b/c you're mixing tabs and spaces w/out benefit. I intentionally included some scope in my preferred way of doing this to show that you use tabs to get to the same indentation level and then spaces the rest of the way (again, so that changing the width of a tab doesn't break alignment).

Your way doesn't align things at all, which is fine I guess, just not suited to my particular flavor of OCD :D

zakird:
I agree with Pat here. Aligning things when using tabs doesn't really make sense and I believe we should just stick with the two tabs. Again, this hasn't caused us any issues so I see no reason to change it.

zakird:
I'm closing this PR. While I'm fine with having a clang format check, we do not want to change the style of this project at this time and a new PR should be opened that makes this single modification.

PR #384, opened 2016-10-23
allow large max line length and fatal when that fails

zakird:


paul-pearce:
This is wrong.

-fgets enforce null termination
-In the error case, fgets will read up to `MAX_LINE_LENGTH - 1` characters, then null terminate
-`strlen(line)` is `MAX_LINE_LENGTH - 1`
-The error check would need to be `strlen(line) >= (MAX_LINE_LENGTH - 1)`

But even if you fix that, it's sub-optimal. Now your max line length is MAX_LINE_LENGTH - 2 because a line that fits exactly will hit the error condition.

Just check that `line[strlen(line) - 1] == '\n'` and be done. You probably want to ensure that `strlen(line) > 0` before doing that.

paul-pearce:
Line 187 has extra whitespace.

Line 189 reports the wrong value in the error msg (should be -1)

PR #386, opened 2016-11-07
Only log to syslog when enabled

### LINKED ISSUE #385: ZMap does not respect --disable-syslog
ZMap always logs to syslog, regardless of if it's enabled or not. If syslog is enabled, ZMap will make an `openlog` call to set default settings. If it is not enabled, ZMap skips the `openlog` call, but still calls `va_syslog`, etc.

dadrian:
Fixed #385


zakird:
LGTM

PR #396, opened 2016-12-06
Handle invalid zblacklist input, arg to toggle printing

### LINKED ISSUE #381: zblacklist --no-duplicate-checking=False filters all invalid lines as duplicates for line numbers > 1 (accessing uninitialized memory)
> $ cat test_input 
> 1.2.3.4
> not_valid_ip_address
> 8.8.8.8
> $ cat test_input | zblacklist -b /etc/zmap/blacklist.conf 
> 1.2.3.4
> 8.8.8.8
> $ cat test_input | zblacklist -b /etc/zmap/blacklist.conf --no-duplicate-checking 
> 1.2.3.4
> not_valid_ip_address
> 8.8.8.8

The root cause here is using`addr` here https://github.com/zmap/zmap/blob/master/src/zblacklist.c#L198 after a failed `inet_aton` here https://github.com/zmap/zmap/blob/master/src/zblacklist.c#L194. At this point the value of `addr` is uninitialized. This just happens to end up (under ubuntu 14.04) taking on the value from the previous iteration, resulting in a duplicate hit.

The reason this only applies for line numbers > 1 is the `addr`needs something to be a duplicate for.

`addr` is extensively used after the `inet_anon` call.  If we want to keep invalid lines, I think the right fix here is immediately output invalid lines if `-b` is specified and short circuiting the loop, or immediately shirt circuiting the loop if `-w`  with no output.

paul-pearce:
Closes #381

Previously we were reading an uninitialized in_addr struct resulting
in different behavior depending on unrelated options.

We now short circuit when we hit an invalid entry (preventing reading
uninitialized memory, and control printing with a new argument.

I also renamed the existing conf variable from ignore_errors to ignore_blacklist_errors, to match convention everywhere else.

### DIFF
--- a/src/zblacklist.1
+++ b/src/zblacklist.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "ZBLACKLIST" "1" "September 2015" "zmap v2.1.1" "zblacklist"
+.TH "ZBLACKLIST" "1" "December 2016" "ZMap" "zblacklist"
 .
 .SH "NAME"
 \fBzblacklist\fR \- zmap IP blacklist tool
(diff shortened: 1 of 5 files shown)


zakird:
Going to leave this to @paul-pearce until tests pass

dadrian:
I split this into two separate commits, since you're actually changing two things. You're still marked as author. Assuming this passes tests, I'll rebase onto master.

PR #397, opened 2016-12-06
Set default number of sender threads to 1 (was num_cores - 2)

### LINKED ISSUE #374: zmap defaults number of senders to (num_cores - 2), not 1 as documented
zmap help text:

>   -T, --sender-threads=n        Threads used to send packets  (default=`1')

This text is generated here: https://github.com/zmap/zmap/blob/master/src/zopt.ggo.in#L169

However the zmap.c does not honor the default and instead defaults to num_cores - 2 if the option is not set:

https://github.com/zmap/zmap/blob/master/src/zmap.c#L757

I'm happy to fix this with a PR but I'm not sure what the right behavior is. Should the default be 1 or num_cores - 2?

paul-pearce:
This closes #374

### DIFF
--- a/src/zmap.c
+++ b/src/zmap.c
@@ -757,12 +757,7 @@ int main(int argc, char *argv[])
 	if (args.sender_threads_given) {
 		zconf.senders = args.sender_threads_arg;
 	} else {
-		int num_cores = sysconf(_SC_NPROCESSORS_ONLN);
-		zconf.senders = max_int(num_cores - 1, 1);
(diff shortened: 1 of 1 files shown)


paul-pearce:
As an aside: This is got to be one of the first PR's that resulted in a LoC reduction!

zakird:
LGTM

dadrian:
Merged in c04bc6d060f0ce1f198ff7eda6bf1a685d3b8efa

Issue #364, opened 2016-08-23
`zmap -V` should report something other than the latest prior release when built from master


paul-pearce:

paul-pearce:
Resolved with 9e627e1

PR #403, opened 2016-12-28
Replace version string with git commit information for non-release builds

paul-pearce:
Switching back to the classic version can be done by changing the ZMAP_VERSION variable in /CMakeLists.txt to a version number and calling 'make'.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,18 +11,6 @@ option(WITH_WERROR "Build with -Werror" OFF)
 option(WITH_PFRING "Build with PF_RING ZC for send (10 GigE)" OFF)
 option(FORCE_CONF_INSTALL "Overwrites existing configuration files at install" OFF)
 
-if("${ZMAP_VERSION}" STREQUAL "DEVELOPMENT")
-    set(GIT_CMD "git")
-    set(GIT_ARGS "log" "-n" "1" "--pretty=format:%h - %ad")
-    execute_process(COMMAND ${GIT_CMD} ${GIT_ARGS}
-        RESULT_VARIABLE GIT_RESULT
-        OUTPUT_VARIABLE GIT_COMMIT)
-    if (GIT_RESULT)
-        set (GIT_COMMIT "UNKNOWN")
-    endif()
-    set(ZMAP_VERSION "Development Build. Commit ${GIT_COMMIT}")
-endif()
-
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
     set(USING_CLANG "YES")
 else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
 project(ZMAP C)
-set(ZMAP_VERSION 2.1.1)
+set(ZMAP_VERSION DEVELOPMENT) # Change DEVELOPMENT to version number for release
 
 option(WITH_REDIS "Build with support for Redis DB" OFF)
 option(WITH_MONGO "Build with support for MongoDB" OFF)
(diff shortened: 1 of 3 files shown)


paul-pearce:
This resolves #364.

paul-pearce:
Sample output:
```
$zmap -V
zmap Development Build. Commit efc1d77 - Tue Dec 27 16:33:03 2016 -0800
```

src/CMakeLists.txt
```diff
@@ -143,10 +143,19 @@ else()
     set(ZTESTSOURCES ${ZTESTSOURCES} recv-pcap.c)
 endif()
 
-configure_file(topt.ggo.in ${CMAKE_BINARY_DIR}/src/topt.ggo @ONLY)
-configure_file(zbopt.ggo.in ${CMAKE_BINARY_DIR}/src/zbopt.ggo @ONLY)
-configure_file(zitopt.ggo.in ${CMAKE_BINARY_DIR}/src/zitopt.ggo @ONLY)
-configure_file(zopt.ggo.in ${CMAKE_BINARY_DIR}/src/zopt.ggo @ONLY)
+# Set configure time zmap version
+configure_file(topt.ggo.in topt.ggo @ONLY)
```
dadrian: Why did you remove the `${CMAKE_BINARY_DIR}`? You probably broke out of source builds, which will make package maintainers upset.

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
 project(ZMAP C)
-set(ZMAP_VERSION 2.1.1)
+set(ZMAP_VERSION DEVELOPMENT) # Change DEVELOPMENT to version number for release
```
dadrian: We should make this an `option`, instead of requiring editing source.

src/CMakeLists.txt
```diff
@@ -143,10 +143,19 @@ else()
...
+configure_file(zbopt.ggo.in zbopt.ggo @ONLY)
+configure_file(zitopt.ggo.in zitopt.ggo @ONLY)
+configure_file(zopt.ggo.in zopt.ggo @ONLY)
+# Additional ggo.in's should be added here and CMakeVersion.txt
+
+# This sets a *build* time dependency that checks git
+if("${ZMAP_VERSION}" STREQUAL "DEVELOPMENT")
```
dadrian: This should test against the option. See earlier comment.

src/CMakeVersion.txt
```diff
@@ -0,0 +1,14 @@
...
+set(GIT_ARGS "log" "-n" "1" "--pretty=format:%h - %ad")
+execute_process(COMMAND ${GIT_CMD} ${GIT_ARGS}
+    RESULT_VARIABLE GIT_RESULT
+    OUTPUT_VARIABLE GIT_COMMIT)
+    if (GIT_RESULT)
+        set (GIT_COMMIT "UNKNOWN")
+    endif()
+set(ZMAP_VERSION "Development Build. Commit ${GIT_COMMIT}")
+
+configure_file(topt.ggo.in topt.ggo @ONLY)
```
dadrian: Same comment about out of source builds. Also, have you tested what happens when one of the `.ggo.in` files is modified? Also, why not just write a shell script and use `add_custom_command`?

CMakeLists.txt
paul-pearce: This was based on a discussion with @zakird. I'm fine switching this to an option instead if that's what's desired.

src/CMakeLists.txt
paul-pearce: Point taken.

src/CMakeVersion.txt
paul-pearce: 1) This was not checked against out of source builds (which don't seem to be supported by the project at all)

2) Modifying the .ggo.in files is fine? What is your concern? 

3) Doing this inside the build system seemed preferable to a shell script, especially given that shell scripts aren't used elsewhere in the project.

CMakeLists.txt
paul-pearce: More context: I think the plan was that release versions would be git branches with the variable replaced with the release number. That way even when building from source you'd get the version rather than the dev string. Of course this could also be done by the branch setting the dev mode 'off' by default.

src/CMakeVersion.txt
dadrian: 1. Point taken.
2. I wanted to make sure the dependencies were still picked up correctly, and that this was triggered after a modification.
3. Seemed easier to implement in an easier language, especially being able to explicitly note dependency chains in `add_custom_command`. See (2)

CMakeLists.txt
dadrian: oooo thats clever

src/CMakeVersion.txt
paul-pearce: 1) I did my best to patch this up, but since out of source builds are broken elsewhere, it's not fully tested. But I think it will work if we fix the top-level issues.

2) Ah. Yes, that's not impacted since we're doing per-build generation when in development mode.

3) Let's decouple the cmake 'add_custom_WHAT` from what it invokes (since both command and target can invoke either).

Between `add_custom_target` and `add_custom_command`, target seems better since you want this invoked every build. That's what `add_custom_target` is built for. What dependency would you specify for `add_custom_command`? Seems like we'd need to log the commit in some file somewhere and add per-build hooks to check the value. Since both require per-build hooks, skipping writing state to a file somewhere seems preferable.

RE what to invoke: Since we're replicating what the build system is already doing (replacing a cmake variable in a file) reusing the build system seemed preferable. I also like not having bash scripts laying around in the project. Do you care enough to request this be scrapped and replaced with a bash script?  (The current solution seems complete, not withstanding style comments.)

dadrian: If it works, shipit.

zakird:
@paul-pearce to clarify, this handles getting the git commit when make is run correct, not just cmake?

paul-pearce:
Yes.

PR #414, opened 2017-03-23
Fixed typo in monitor.c

amanelis:


zakird:
Good catch. Thanks for the correction.

PR #419, opened 2017-04-09
Improved ICMP validation

zakird:
This adds validation to the ICMP validation to the RTT scan, and also extends validation to use both ICMP id and sequence number.

### DIFF
--- a/src/probe_modules/module_icmp_echo_time.c
+++ b/src/probe_modules/module_icmp_echo_time.c
@@ -6,8 +6,8 @@
  * of the License at http://www.apache.org/licenses/LICENSE-2.0
  */
 
-// probe module for performing ICMP echo request (ping) scans that allows calculation
-// of RTT
+// probe module for performing ICMP echo request (ping) scans that allows
+// calculation of RTT
 
 #include <stdlib.h>
 #include <stdio.h>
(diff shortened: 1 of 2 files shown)


zakird:
@dadrian @paul-pearce: tagging both of you given that this is a fairly core piece of ZMap.

dadrian:
Were we just not setting ICMP seqno previously?

zakird:
Nope!

src/probe_modules/module_icmp_echo_time.c
```diff
@@ -35,184 +35,193 @@ struct icmp_payload_for_rtt {
...
+			icmp_header->icmp_code,
+			ntohs(icmp_header->icmp_cksum),
+			ntohs(icmp_header->icmp_id),
+			ntohs(icmp_header->icmp_seq));
+	fprintf_ip_header(fp, iph);
+	fprintf_eth_header(fp, ethh);
+	fprintf(fp, "------------------------------------------------------\n");
 }
 
 static int icmp_validate_packet(const struct ip *ip_hdr,
```
dadrian: Why are there two `icmp_validate_packet` functions that seem largely identical?

paul-pearce: This problem exists across other modules as well. In our copious amounts of spare time we should try to unify and standardize how this is done.

paul-pearce:
The changes lgtm.

Reading this I began wondering why you switched idnum from validation[2] to validation[1]. I then began wondering why you don't use validation[0]. From digging around, it seems like there's no rhyme or reasons why we use any particular validation field across any of the probe modules. We should probably regularize how this works. e.g. starting from validation[0] we assign in the order the fields occur in the packet. 

But that's a broader issue.

zakird:
I was confused by this too.

PR #420, opened 2017-04-10
Extracting IP header from ICMP error message payload

sheharbano:
The following line assumes that sizeof(struct icmp) == 8. It does not correctly extract the original IP header from ICMP error message payload.

struct ip *ip_inner = (struct ip *) &icmp[1];

I replaced the line above with this one:

// ICMP header is 8 bytes long
struct ip *ip_inner = (struct ip*) ((char *)icmp + 8);


zakird:
It looks like this doesn't currently build. Can you take a look?

sheharbano:
Sorry about that. Can you please try building with the suggested change.

It should be 
struct ip *ip_inner = (struct ip*) ((char *) icmp_h+8);

rather than
struct ip *ip_inner = (struct ip*) ((char *)icmp + 8);

dadrian:
@sheharbano It doesn't look like this merges or compiles. Can you update this PR?

sheharbano:
Updated. I just tried, successfully built on my end.

zakird:
This looks like it still unfortunately conflicts with master and cannot be merged.

PR #424, opened 2017-04-14
Continue: Extracting IP header from ICMP error message payload #420

sheharbano:
I reverted my changes and rebased my zmap fork to continue working on the upstream zmap version (I think this also led to the PR getting closed automatically -- I didn't manually close it). Then I changed the ip_inner line and make'd zmap successfully. Hope this works now.

### DIFF
--- a/src/probe_modules/module_icmp_echo.c
+++ b/src/probe_modules/module_icmp_echo.c
@@ -111,7 +111,7 @@ static int icmp_validate_packet(const struct ip *ip_hdr,
 			sizeof(struct ip)) > len) {
 			return 0;
 		}
-		struct ip *ip_inner = (struct ip*) ((char *) icmp_h+8);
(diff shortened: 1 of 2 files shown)


zakird:
This doesn't quite look like the right of making this change in C. However, I also can't quite tell what you're trying to fix offhand. Can you explain at a higher level what problem is PR is trying to fix?

sheharbano:
It's the same as https://github.com/zmap/zmap/pull/420 which got closed. The following line assumes that sizeof(struct icmp) == 8. It does not correctly extract the original IP header from ICMP error message payload.

struct ip *ip_inner = (struct ip *) &icmp[1];

I replaced the line above with this one:

// ICMP header is 8 bytes long
struct ip ip_inner = (struct ip) ((char *)icmp + 8);

Issue #407, opened 2017-01-23
Verify / Fix out of source builds


paul-pearce:
It's unclear if out of source builds are currently working. Test, fix as necessary.

zakird:
Is this still an issue?

dadrian:
I don't know if it builds out of source or not, but regardless we should do an out of source build on TravisCI to make sure it keeps working.

ogasser:
Out of source builds still fail:
```
CMake Error: Error processing file: CMakeVersion.txt
make[2]: *** [src/CMakeFiles/git_versioning.dir/build.make:57: src/CMakeFiles/git_versioning] Error 1
make[1]: *** [CMakeFiles/Makefile2:336: src/CMakeFiles/git_versioning.dir/all] Error 2
make: *** [Makefile:150: all] Error 2
```

paul-pearce:
Fixed by #425

PR #425, opened 2017-04-28
Restore out of source builds

paul-pearce:
Addresses #407

We now once again support out of source builds. I added notes on performing an out of source build to the readme.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,4 +33,3 @@ CMakeCache.txt
 install_manifest.txt
 _CPack_Packages/*
 src/ztests
-build/
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ CMakeCache.txt
 install_manifest.txt
 _CPack_Packages/*
 src/ztests
+build/
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -69,6 +69,16 @@ and installed by running:
   ```
 
 # Miscellaneous notes:
+- Out of source builds can be performed by running:
+  ```sh
+  cmake -H"." -B"build/"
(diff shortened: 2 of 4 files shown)


paul-pearce:
It's worth pointing out that `cmake .` throws a ton of files into the source directory that you can only get rid of by either 1) deleting everything from `.gitignore` or 2) deleting the repo and recloning. It's a bit of a mess. There's no notion of `cmake clean` in cmake, and most recommend out of source builds.

It's worth considering if we want to recommend out of source builds by default. The main drawback is it makes the install instructions less awesome, as the cmake options for out of source builds are undocumented.

paul-pearce:
@zakird nor do I. Should I go ahead and merge or do we think @dadrian might not understand cmake less than us? 

(I also added `build/` to the gitignore so if you follow the readme it doesn't try to commit the build dir.)

dadrian:
Out of source builds are traditionally done by doing `cd /some/random/path; cmake /path/to/zmap`. Does this work?

paul-pearce:
@dadrian, interesting. The (poor) documentation I found indicated otherwise.

Yes, that works too.

paul-pearce:
My rationale for including that is I think we should actually encourage out of source builds. 

It's a bit crazy that once you type `cmake .` you can never go back without deleting the entire source repo and recloning.

dadrian:
You can `git clean -dfx` and you're fine. You don't have to reclone.

There's no reason to arbitrarily add random subdirectories to our .gitignore. If we want to encourage out of source builds, we should have the examples build to an adjacent directory to the Git repo, and not a magic subdirectory.

Issue #430, opened 2017-05-23
Impossible to get non-successful packets


zakird:
Now that output-module doesn't change output-filter, it's impossible to get RST packets out of ZMap.

paul-pearce:
This is related to the fix for issue #377 and pr #394

Issue #377, opened 2016-09-09
Default --output-module overrides --output-filter


paul-pearce:
The default output filter enforces success = 1, regardless of what you pass into --output-filter.

This yields 0 results:
`
zmap -p 80 --probe-module="tcp_synscan" --output-fields=* --output-filter="success = 0" 
`

Whereas this yields results:

`zmap -p 80 --probe-module="tcp_synscan" --output-fields=* --output-module="csv" --output-filter="success = 0" 
`

paul-pearce:
We think this is not resolved with PR #394, and it introduced poor additional behavior. Reopening this until we sort things out.

paul-pearce:
Fixed, we think

PR #432, opened 2017-05-26
fixed output filter nonsense

zakird:


paul-pearce:
lgtm. This addresses #430 and #377

Issue #433, opened 2017-05-27
Recent gateway buffer changes introduced problems detecting GW mac


paul-pearce:
PR #428, which aimed to resolve issue #421 has introduced issues detecting gateways under some configuration. (e.g., my groups scanning environment.) Investigation is ongoing.

paul-pearce:
It looks like PR #428 never worked. It passes in sizeof(buf) which used to be 8K and is now 8 (as we went from a static array to a ptr) @ https://github.com/zmap/zmap/blob/08a6e677c259dec11e8230ff5707b83ede9706f6/src/get_gateway-linux.h#L112. Further, the failure check at https://github.com/zmap/zmap/blob/08a6e677c259dec11e8230ff5707b83ede9706f6/src/get_gateway-linux.h#L113 also never worked, since you're checking if an unsigned int is < 0. The only reason gateway failures ever occured was a followon check failed to parse all 0s.

paul-pearce:
Fixed.

Issue #421, opened 2017-04-12
Bug of could not detect GW MAC address


dragonkid:
I did have the MAC address of gateway in my ARP cache table, and I also tried to run `arp <gateway_ip>`. But the error message continued.

```
Apr 12 14:56:35.997 [FATAL] zmap: could not detect GW MAC address for 10.11.11.254 on eth0. Try setting default gateway mac address (-G), or run "arp <gateway_ip>" in terminal.
```

I recognized that I have a large ARP cache table with 117 records, but the method of `get_hw_addr` in `get_gateway-linux.h:95` only read 8192 bytes of it. When I expand the buf size to 16384, the error disappeared.

I think this was not the right way to fix this error, any suggestions?

dadrian:
Arguably, we should extend `read_nl_socket` to handle arbitrary length responses, but I don't particularly want to introduce dynamic memory management. I don't think I would be opposed to a PR that doubled the size of that buffer.

@zakird thoughts?

zakird:
I don't think this is worth it. Let's just make it like like 64KB and move on.

PR #434, opened 2017-05-27
Fix gateway mac detection on linux

paul-pearce:
unsigned ints cant be less than 0, and the sizeof(ptr) != sizeof(static array)


paul-pearce:
Fixes / properly fixes some combination of #433 and #428 and #421

PR #437, opened 2017-06-15
Add --ignore-blacklist-errors zmap flag

### LINKED ISSUE #436: Blacklisted host that does not resolve to an IP causes error.
https://github.com/zmap/zmap/blob/master/lib/blacklist.c#L131

getaddrinfo will return non-zero for a host that resolves to no IP. The program promptly throws up its hands and quits. A better behavior is to just ignore that entry.

cdzombak:
`zblacklist` uses the flag `--ignore-blacklist-errors` to indicate that blacklist parsing/resolution errors should be ignored. `zmap` uses `--ignore-invalid-hosts`.

This PR adds `--ignore-blacklist-errors` to `zmap`; it's equivalent to `--ignore-invalid-hosts`, but these tools should allow using similar flag names for similar functionality.

While updating the manpage, I was disappointed that regenerating the `roff` and `html` files was a manual process, so I added a new target which regenerates manpages (assuming [`ronn`](https://github.com/rtomayko/ronn) is available). That target does not run by default, since many builders won't have `ronn` installed.

I also added a few generated files to `gitignore`, and as long as I was editing INSTALL documented how to install `zmap` to a custom location.

- closes #436


zakird:
I'm not quite sure if I follow the logic to change ZMap instead of
ZBlacklist. ZMap is the far more mature of the two projects, and the only
one that has been semantically versioned and released in OSes. I would much
prefer to see ZBlacklist change to match ZMap and not vice-versa.

dadrian:
It looks like you added two files to .gitignore but didn't remove them from the repository.

cdzombak:
> I'm not quite sure if I follow the logic to change ZMap instead of
ZBlacklist. ZMap is the far more mature of the two projects, and the only
one that has been semantically versioned and released in OSes. I would much
prefer to see ZBlacklist change to match ZMap and not vice-versa.

I see your point, but I'm thinking about the names of these options. Without looking at the docs, it's clear to me what "ignore blacklist errors" means. It's not clear what "ignore invalid hosts" means. Looking forward, I think it makes sense to choose whichever name is clearer.

> I would also add that I'd prefer to just change zblacklist than to add convulution
to the flags that ZMap uses. One of the major struggles with nmap is that
it has a very confusing set of command line flags, and I'd like to try to avoid
that situation if at all possible.

Makes sense. If the next zmap/zblacklist version is 3.0 (ie. a major update), simply removing whichever option we choose to drop is a good choice.

cdzombak:
> It looks like you added two files to .gitignore but didn't remove them from the repository.

Those files (`ztopt.c`/`.h`) have never existed in the repo; they were just generated locally.

zakird:
I'm fine with swapping to the better name. However, if we're making an active choice to change the name of the ZMap option, we need to throw a deprecation log warning when someone uses the "wrong" option, such that it can be removed in 4.0. Please also create an associated issue for that milestone.

cdzombak:
> I'm fine with swapping to the better name. However, if we're making an active choice to change the name of the ZMap option, we need to throw a deprecation log warning when someone uses the "wrong" option, such that it can be removed in 4.0.

@zakird https://github.com/zmap/zmap/pull/437/commits/9746330540552e6bb383cb1f352cb8f05c70aa44 adds such a deprecation warning. I'm happy to adjust this until you're satisfied with the message & its implementation.

> Please also create an associated issue for that milestone.

I'll do that once this PR is accepted.

src/zmap.c
```diff
@@ -523,10 +523,14 @@ int main(int argc, char *argv[])
 		zconf.output_filter_str = args.output_filter_arg;
 		log_debug("filter", "will use output filter %s", args.output_filter_arg);
 	}
+	// --ignore-invalid-hosts is deprecated and will be removed in zmap 4.0
+	if (args.ignore_invalid_hosts_given) {
+		log_warn("zmap", "--ignore-invalid-hosts is deprecated and will be removed in zmap 4.0. Use --ignore-blacklist-errors instead to ignore invalid blacklist/whitelist entries.");
+	}
+	zconf.ignore_invalid_hosts = args.ignore_invalid_hosts_given || args.ignore_blacklist_errors_given;
```
zakird: It might be worth changing this throughout to be consistent.

zakird:
Curious, why do some man files end up with May 2017 instead of June 2017?

cdzombak:
> Curious, why do some man files end up with May 2017 instead of June 2017?

`ronn` uses "The file mtime is used when no date is given, or the current time when no file is available." So … perhaps because I cloned this repo in May?

src/zmap.c
cdzombak: I considered this, but `ignore_invalid_hosts` is the parameter name used by the blacklist library in all its functions, so this'll be a fairly pervasive change. If that's worth the noise, I'm happy to do it.

src/zmap.1
```diff
@@ -212,8 +219,8 @@ Minimum hitrate that scan can hit before scan is aborted
 Comma\-separated list of cores to pin to
 .
 .TP
-\fB\-\-ignore\-invalid\-hosts\fR
-Ignore invalid hosts in whitelist/blacklist file
+\fB\-\-ignore\-blacklist\-errors\fR
+Ignore invalid entries in whitelist/blacklist file
```
zakird: It might be worth it to be explicit here. IIRC, this doesn't ignore malformed entries, it just allows names that can't be resolved.

src/zmap.c
```diff
@@ -523,10 +523,14 @@ int main(int argc, char *argv[])
 		zconf.output_filter_str = args.output_filter_arg;
 		log_debug("filter", "will use output filter %s", args.output_filter_arg);
 	}
+	// --ignore-invalid-hosts is deprecated and will be removed in zmap 4.0
+	if (args.ignore_invalid_hosts_given) {
+		log_warn("zmap", "--ignore-invalid-hosts is deprecated and will be removed in zmap 4.0. Use --ignore-blacklist-errors instead to ignore invalid blacklist/whitelist entries.");
```
zakird: This is stupidly pedantic, but I'd prefer to see `ZMap` not `zmap`.

src/zmap.c
zakird: sgtm

src/zmap.c
cdzombak: Fixed in 5ea24dc37deb45a395a5e513c8962f6238c3d1f7

src/zmap.1
cdzombak: This option does seem to make ZMap ignore malformed entries like `123.456./24`. Added clarification in dba82b5223df7e77ce6d338716342b967c12d66a.

PR #438, opened 2017-06-15
Make check_manfile.py verify ztee, zblacklist, ziterate manpages

cdzombak:
Travis runs `check_manfile.py` to verify that the man pages for ZMap etc. include all the options listed in the tools' `gengetopt` files. However, this script previously did not verify the man pages for `ztee`, `zblacklist`, or `ziterate`.

This PR enforces this consistency check for those tools' man pages, and it corrects the inconsistencies that existed.


scripts/check_manfile.py
```diff
@@ -1,21 +1,36 @@
...
+
+for ggo, ronn in checks:
+    options = []
+    with open("src/" + ggo) as fd:
+        for l in fd:
+            if l.startswith("option "):
+                option = l.split()[1].lstrip('"').rstrip('"')
+                options.append(option)
+
+    man = open("src/" + ronn).read()
```
dadrian: Pretty sure this should still be within a `with`, but also who really cares.

PR #442, opened 2017-07-04
add support for elasticsearch in output modules

shjalayeri:
The elk output module enables the possibility of writing scan outputs to a elasticsearch index.


dadrian:
Thank you for putting this output module together. We're happy that the output module framework is able to be used in this way. However, we (the ZMap team) don't want to maintain this module long term, and have found in our experience it's not reliable to pipe ZMap directly into anything except a stream provider. Thanks!

PR #443, opened 2017-07-24
Allow make_packet to change packet length from module-wide default

paul-pearce:


paul-pearce:
There's a followup PR coming to enable multiple DNS questions which utilizes the feature introduced in this PR.

src/probe_modules/module_bacnet.c
```diff
@@ -73,7 +73,8 @@ int bacnet_init_perthread(void *buf, macaddr_t *src,
 }
 
 int bacnet_make_packet(void *buf, ipaddr_n_t src_ip, ipaddr_n_t dst_ip,
-				uint32_t *validation, int probe_num, UNUSED void *arg)
+				uint32_t *validation, int probe_num, UNUSED void *arg,
+				UNUSED int *len)
```
zakird: It might make the most sense to have length right after buffer given that that's what's being described.

zakird: why is length an int and not a `size_t`?

src/send.c
```diff
@@ -331,14 +331,14 @@ int send_run(sock_t st, shard_t *s)
 			uint32_t src_ip = get_src_ip(curr, i);
 		  	uint32_t validation[VALIDATE_BYTES/sizeof(uint32_t)];
 			validate_gen(src_ip, curr, (uint8_t *)validation);
+			int length = zconf.probe_module->packet_length;
```
zakird: This should probably be a `size_t`.

src/probe_modules/module_bacnet.c
paul-pearce: Placing it next to `buf` seemed "disruptive" given that it's essentially never used. I followed the lead from `arg` which is also never used, and placed it at the end.

But I'm not opposed. Do you want me to make the change?

src/send.c
paul-pearce: Agreed. The existing code used `int` so I just went with it.

paul-pearce:
Re safety check, agreed.

src/probe_modules/module_bacnet.c
zakird: Let's see what @cdzombak and @dadrian think. I'm not sure what makes sense to be honest.

paul-pearce: When they weigh in, input on a name would also be appreciated. I'm going to go with buf_len otherwise.

cdzombak: With the caveat that I'm quite unfamiliar with this codebase: I agree with @zakird here. Though that length may be infrequently used, keeping the buffer pointer and its length as close together as possible should reduce potential for errors.

Re; the name: `buf_len` seems perfectly self-explanatory to me, especially if it's located adjacent to `buf`.

paul-pearce:
PR has been updated to:
1) Change everything to a size_t
2) Rename the variable to buf_length
3) Add a safety check to not exceed max packet size.

I've not changed the arg order since there was not yet consensus.

src/probe_modules/module_bacnet.c
dadrian: `buf_len` is okay, but it did take me a moment to realize that it was a pointer, because I'm used to lengths being not pointers-to-length.

src/send.c
```diff
@@ -331,14 +331,15 @@ int send_run(sock_t st, shard_t *s)
 			uint32_t src_ip = get_src_ip(curr, i);
 		  	uint32_t validation[VALIDATE_BYTES/sizeof(uint32_t)];
 			validate_gen(src_ip, curr, (uint8_t *)validation);
-			zconf.probe_module->make_packet(buf, src_ip, curr,
+			size_t length = zconf.probe_module->packet_length;
+			zconf.probe_module->make_packet(buf, &length, src_ip, curr,
 					validation, i, probe_data);
+			assert(length <= MAX_PACKET_SIZE);
```
zakird: can we make this a log_fatal instead of an assert?

paul-pearce: What are the project guidelines on the use of each?

PR #444, opened 2017-07-25
Support for multiple DNS questions

paul-pearce:
This adds support for multiple DNS questions via the syntax:
"A,example.com;AAAA,www.example.com"

To use this you must set --probes to match the number of questions
and --output-filter="" (or something non default) to avoid
filter_duplicates


paul-pearce:
Part of this PR is a retab to get module_dns.c to conform to the project standards. The retab is its own commit, so you may wish to diff by commit instead of the PR as a whole.

PR #451, opened 2017-08-31
Don't squash status_updates_file descriptor names

### LINKED ISSUE #431: zmap --status-updates-file param seems unstable
zmap -u param seems unstable,I use the following command,`zmap -p 80 --output-fields=* -b -B 30M -l zmap-log.log -m zmap-metadata.json -u zmap-status.csv --quiet`, it often occured the error 'could not open output file zmap-status.csv',I need to run many times to succeed.I have used the sudo

dadrian:
Fixes #431

### DIFF
--- a/src/monitor.c
+++ b/src/monitor.c
@@ -10,6 +10,7 @@
 
 #include "monitor.h"
 
+#include <assert.h>
 #include <errno.h>
 #include <math.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


src/monitor.c
```diff
@@ -379,7 +380,8 @@ static inline void check_max_sendto_failures(export_status_t *exp) {
 
 void monitor_init(void) {
 	if (zconf.status_updates_file) {
-		f = init_status_update_file(zconf.status_updates_file);
+		status_fd = init_status_update_file(zconf.status_updates_file);
+		assert(status_fd);
```
zakird: This should probably be a log fatal.

zakird: This should probably be a log fatal.

dadrian: It does in `init_status_update_file`

PR #461, opened 2017-09-06
Run GNU `indent -kr -i8 src/send.c`

dadrian:
This file was totally screwed up.

### DIFF
--- a/indent.sh
+++ b/indent.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+PARAM="-npro -kr -i8 -ts8 -sob -l80 -ss -ncs -cp1"
+CMD="indent"
+if [ $(uname -s) = "Darwin" ]; then
+	CMD="gindent"
+fi
+
+RES=`$CMD --version | cut -d' ' -f3`
+if [ "$RES" = "" ]; then
+	exit 1
+fi
+V1=`echo $RES | cut -d'.' -f1`
+V2=`echo $RES | cut -d'.' -f2`
+V3=`echo $RES | cut -d'.' -f3`
+
+if [ $V1 -gt 2 ]; then
+  PARAM="$PARAM -il0"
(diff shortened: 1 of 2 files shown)


paul-pearce:
Given zakir's comment, would a retab in vim behave better? that's what I've been doing when I run into this problem on this project.

dadrian:
Blegh Kernel style guide doesn't even specify what your'e supposed to do here.

dadrian:
Alright, so I'm getting real tired of not having a programmatic way of fixing style, given then ZMap uses a different style than all of the other code that I write in the same language family. I can't be the only person that has to reconfigure (or change) editors whenever I'm writing code for ZMap. It's also clear we've done a terrible job at enforcing anything close to the style guide, and this makes code annoying to read and write, and error-prone.

Nominally, we currently use kernel style. However, when I ran the same shell script that runs the indenter for the Linux kernel tree (https://raw.githubusercontent.com/torvalds/linux/master/scripts/Lindent), it's pretty clear that we aren't quite doing that. We're close, but we're not quite there.

In my opinion, this leaves us with three options:

1. Switch to actual kernel style. Run the indent script on all files. I have a script that works on Linux and OS X with `brew install gnu-indent` in the current state of this PR. Use this as our style going forward.
2. Switch to some style supported by clang-format out of the box (e.g. Chromium). I don't recommend this, we're not writing C++. This would be a good time to switch from tabs to spaces, if we so desired.
3. Figure out whatever style we happen to be writing in, and write a script that uses `indent` or `clang-format` to enforce it. This seems silly.

I highly recommend we take option (1).

cdzombak:
We should note this new style guideline in README (or in CONTRIBUTING if we eventually add one).

dadrian:
I'll follow up with another PR to apply the style to everything, and update README.

PR #463, opened 2017-09-07
Apply kernel style to all code

dadrian:
Update INSTALL.md to explain style and indent.sh.


dadrian:
Sigh. Indent does weird alignment with pointers to typedef'd types in function declarations, and the option to fix this `--pointer-align-right` doesn't exist in most versions. Linux kernel doesn't really seem to encounter this because they explicitly use the `struct` keyword instead of just typedef'ing everything. Looking into `clang-format`.

dadrian:
OK, I have attempted this again, except with `clang-format`.

PR #464, opened 2017-09-11
Simplify sharding implementation

### LINKED ISSUE #410: Ziterate doesn't shard correctly if num_shards is not a factor of number of IPs
If num_shards is not a factor of number of IPs (that is, p-1), then ziterate doesn't seem to cycle through all the IPs. This was run on v1.1.2-849-g967e395 on debian.

```
zack@debian:~/pkg/zmap$ cat /tmp/w
0.0.0/24
zack@debian:~/pkg/zmap$ ziterate -e 10 -w /tmp/w | wc -l
255
zack@debian:~/pkg/zmap$ ziterate -e 10 --shards=3 --shard=0 -w /tmp/w | wc -l
29
zack@debian:~/pkg/zmap$ ziterate -e 10 --shards=3 --shard=1 -w /tmp/w | wc -l
28
zack@debian:~/pkg/zmap$ ziterate -e 10 --shards=3 --shard=2 -w /tmp/w | wc -l
28
```

dadrian:
Instead of doing things where each shard "skips" N elements each
iteration, just directly divide up the exponent space into N chunks, and
calculate the start and stop point for each. The visualization for this
makes the shards look like slices of a pizza that is the cyclic group,
rather than the weird stuff from the Zippier ZMap paper.

Tests: There were the original tests/test_sharding.py and
test_big_group.sh. These still pass. I added another tests/test-shard.sh
that does a few more tests.

Fixes #410


src/shard.c
```diff
@@ -16,78 +16,113 @@
...
+	// ZMap (set with --shards=N, --shard=n) may have several subshards, if
+	// ZMap is being ran multithreaded (set with --sender-threads=T).
+	//
+	// Total number of subshards is S = N*T. Subshard ID's range from [0,
+	// N*T).
+	assert(num_shards > 0);
+	assert(num_threads > 0);
+	assert(shard_idx < num_shards);
+	assert(thread_idx < num_threads);
+	uint32_t num_subshards = (uint32_t)num_shards * (uint32_t)num_threads;
```
zakird: This looks like a potential overflow.

src/shard.c
```diff
@@ -16,78 +16,113 @@
...
+	uint64_t num_elts = cycle->order;
```
zakird: what's an `elt`?

src/shard.c
```diff
@@ -16,78 +16,113 @@
...
+	assert(num_subshards < num_elts);
+	assert(num_subshards < max_total_targets);
```
zakird: Does this get prevented somewhere else?

src/shard.h
```diff
@@ -9,10 +9,13 @@
 #ifndef ZMAP_SHARD_H
 #define ZMAP_SHARD_H
 
+#include <signal.h>
```
zakird: huhwut?

zakird:
./src/ziterate 141.212.121.0/24 --shards=2 --seed=1 --shard=1 | wc -l
64

zakird:
```./src/ziterate 141.212.121.0/24 141.212.122.0/24 --shards=2 --seed=1 --shard=0 | wc -l```
128

```./src/ziterate 141.212.121.0/24 141.212.122.0/24 --shards=2 --seed=1 --shard=1 | wc -l```
109

**Comment by zakird** [comment]
Actually, give me a moment, I think this might be a build issue.

File: src/shard.c (hunk 1 above)
Comment (dadrian) [reply src/shard.c]: These are both upcasted uint8_t's, so no overflow.

File: src/shard.c (hunk 3 above)
Comment (dadrian) [reply src/shard.c]: I believe it gets prevented in main, but I can go back and look.

**Comment by dadrian** [comment]
I actually haven't tested `ziterate` itself yet, it may be getting initialized incorrectly. I'll take a look.

File: src/shard.h (hunk 4 above)
Comment (dadrian) [reply src/shard.h]: I thought I was going to need `sig_atomic_t`, but ended up not. Will remove include.

**Comment by dadrian** [comment]
@zakird Ready for another look

**Comment by zakird** [comment]
So far so good.

This error message is wrong:
```Sep 14 15:58:33.314 [FATAL] ziterate: With 1 total shards, shard number (1) must be in range [0, 1]```

**Comment by zakird** [comment]
I thought that shards would only be different in size by the number of total shards? Did I get that wrong? How big can the difference be? This looks like a more significant difference:


```
zakir:zmap (dadrian/new-sharding-approach)$ ./src/ziterate 141.212.120.0/24 --blacklist=blacklist  --shards=2 --shard=0 --seed=1 | wc -l
67
zakir:zmap (dadrian/new-sharding-approach)$ ./src/ziterate 141.212.120.0/24 --blacklist=blacklist  --shards=2 --shard=1 --seed=1 | wc -l
61
```

PR #465, opened 2017-09-15
Remove --vpn

dadrian:


dadrian:
I confirmed this did not change total number of results for a heavily populated /24

PR #470, opened 2017-09-16
Fix ICMP handling in probe modules

### LINKED ISSUE #354: Inconsistent ICMP Handling
ICMP handling is quite inconsistent between our probe modules. This might be a sign that we should not be handling this within the probe modules themselves. For example, TCP echo scan just drops all ICMP responses, UDP handles some correctly, but not all, and ICMP module has yet a third set of logic. This is all complicated by the fact that the ICMP sender might not actually be server we tried to send a packet to in the first place.

### LINKED ISSUE #655: Unable to filter for unsuccessful ICMPs
**Describe the bug**

I want to examine ICMP error codes of unsuccessful pings and see how long they took to manifest. I am able to filter on successful ICMP requests but not unsuccessful ones.

**CLI Arguments**

```
$ sudo zmap --probe-module=icmp_echo_time \
            -o result.csv \
            --output-module=csv \
            --output-fields=type,code,repeat,cooldown,icmp_id,seq,sent_timestamp_ts,sent_timestamp_us,dst-raw,classification,success \
            --output-filter="success = 0" \
            192.168.0.0/24
```

I've removed the default 192.168 rule from the boilerplate blacklist:

```bash
$ grep 192.168 /etc/zmap/blocklist.conf
#192.168.0.0/16      # RFC1918: Private-Use
```

If I include ``--verbosity=5`` in the command then the following is printed out:

```
Jun 22 13:14:05.866 [DEBUG] filter: will use output filter success = 0
...
Jun 22 13:14:06.947 [DEBUG] recv: unsuccessful responses will be included in output
```

The above leads me to believe unsuccessful pings will be included in the output.

The above does run without issue and results in:

```
Jun 09 17:14:31.300 [INFO] zmap: completed
```

**Example Target IP**

I am running this on my local network 192.168.0.0/24 for now.

**Expected behavior**

I would expect a CSV with all but the 7 IPv4s in use on my local network to be listed in the results file, with their ICMP error type and the microseconds it took for each failure to occur. Instead I'm given a CSV with just the headers.

**Environment:**
 - OS: Raspbian
 - Version: GitHub HEAD, compiled

zakird:
This PR represents the effort to standardize ICMP handling across all of our probe modules. 

Fixes #655
Fixes #354


zakird:
This branch solely handles ICMP validation and does not output any of the data. However, it's already a _giant_ PR, so we might as well get through this one first.

ajholland:
@zakird I'd be happy to take this over if you'd like, do you have an idea of what still needs to be done? Otherwise, I can just go through and figure out what was already done and proceed from there.

zakird:
If you want to take over and finish it off, that'd be amazing. I haven't had the time to look at this in quite a while. The idea of the PR was originally just to make sure that we included ICMP output consistently in all of our probe modules. This also meant cleaning up how we validated ICMP packets in all of the modules. 

As I started through this, I found that we had a bunch of duplicated code in all of the probe modules and I tried to pull all of these out into helper functions such that our probe modules looked sane. I think that a lot of what's there is OK, but there hasn't been much testing of the code (I wrote all of it on a transatlantic flight without Internet). 

I think that looking through diffs will probably drive you insane given how much has changed. I'd probably just start by getting it to compile (which I think should be pretty easily fixed) and then starting to run some scans and seeing what starts to come out. I'd probably start with the TCP module.

src/probe_modules/module_tcp_synscan.c
```diff
@@ -83,71 +88,111 @@ void synscan_print_packet(FILE *fp, void *packet)
...
+			fs_add_bool(fs, "success", 0);
+		} else { // SYNACK packet
+			fs_add_string(fs, "classification", (char *)"synack",
+				      0);
+			fs_add_bool(fs, "success", 1);
+		}
+		fs_add_null_icmp(fs);
+	} else if (ip_hdr->ip_p == IPPROTO_ICMP) {
+		// tcp
+		fs_add_null(fs, "sport");
```
ajholland: Is adding sport, dport, and classification here **and** in fs_populate_icmp_from_iphdr intentional? If not, I'll fix that.

ajholland:
I did a tcp synscan (port 80) using this branch to see what I'd find. I didn't find any icmp_unreach classifications in a 5%. 

I added an exit(1) condition to validate packet and process packet if the packet wasn't a tcp packet (just for santity checking), and redid a 5% scan. Zmap didn't exit, so no non-tcp packets found. I think my next step is to attempt to send non-tcp packets to a running zmap process and see what happens. 

Any thoughts/am I barking up the wrong tree?

zakird:
Is th pcap filter updated to allow icmp responses?

src/probe_modules/probe_modules.c
```diff
@@ -34,7 +33,6 @@ extern probe_module_t module_bacnet;
 probe_module_t *probe_modules[] = {
     &module_tcp_synscan, &module_tcp_synackscan, &module_icmp_echo,
     &module_icmp_echo_time, &module_udp, &module_ntp, &module_upnp, &module_dns,
-    //&module_tcp_cisco_backdoor,
```
dadrian, line 37: We need to move this to `//examples`

src/probe_modules/packet.h
```diff
@@ -123,14 +158,78 @@ check_dst_port(uint16_t port, int num_ports, uint32_t *validation)
...
+					   uint32_t len)
+{
+	// buf not large enough to contain expected udp header
+	if ((4 * ip_hdr->ip_hl + sizeof(struct icmp)) > len) {
+		return NULL;
+	}
+	return (struct icmp *)((char *)ip_hdr + 4 * ip_hdr->ip_hl);
+}
+
+static inline char *get_udp_payload(const struct udphdr *udp,
```
dadrian, line 207: I'd do uint8_t instead of char, since we think of it as a byte array, not a string.

src/probe_modules/packet.h
```diff
@@ -123,14 +158,78 @@ check_dst_port(uint16_t port, int num_ports, uint32_t *validation)
...
+	if (len < sizeof(struct ether_header)) {
+		return NULL;
+	}
+	return (struct ip *)&packet[sizeof(struct ether_header)];
+}
+
+static inline struct tcphdr *get_tcp_header(const struct ip *ip_hdr,
+					    uint32_t len)
+{
+	// buf not large enough to contain expected udp header
```
dadrian, line 180: ?

src/probe_modules/packet.h
```diff
@@ -123,14 +158,78 @@ check_dst_port(uint16_t port, int num_ports, uint32_t *validation)
...
+	if ((4 * ip_hdr->ip_hl + sizeof(struct tcphdr)) > len) {
+		return NULL;
+	}
+	return (struct tcphdr *)((char *)ip_hdr + 4 * ip_hdr->ip_hl);
+}
+
+static inline struct udphdr *get_udp_header(const struct ip *ip_hdr,
```
dadrian, line 187: Are there UDP options that influence the length of the header?

src/probe_modules/packet.h
```diff
@@ -123,14 +158,78 @@ check_dst_port(uint16_t port, int num_ports, uint32_t *validation)
 	return (((max - min) % num_ports) >= ((to_validate - min) % num_ports));
 }
 
-static __attribute__((unused)) inline uint16_t
+static UNUSED inline uint16_t
```
dadrian, line 161: How is this unused

src/probe_modules/packet.h
```diff
@@ -60,19 +94,20 @@ static inline unsigned short in_icmp_checksum(unsigned short *ip_pkt, int len)
 	return (unsigned short)(~sum);
 }
 
-__attribute__((unused)) static inline unsigned short
+UNUSED static inline unsigned short
```
dadrian, line 97: Why are all of these functions marked UNUSED? Aren't they used?

src/probe_modules/packet.c
```diff
@@ -137,3 +227,13 @@ char *make_ip_str(uint32_t ip)
 	strcpy(retv, temp);
 	return retv;
 }
+
+const char *icmp_unreach_strings[] = {
+    "network unreachable",      "host unreachable",
```
dadrian, line 232: This is...unique formatting.

dadrian: I recognize this is a copy paste, not your fault

src/probe_modules/packet.c
```diff
@@ -127,6 +127,96 @@ void make_udp_header(struct udphdr *udp_header, port_h_t dest_port,
...
+		return PACKET_INVALID;
+	}
+	// find original destination IP and check that we sent a packet
+	// to that IP address
+	uint32_t dest = ip_inner->ip_dst.s_addr;
+	if (!blocklist_is_allowed(dest)) {
+		return PACKET_INVALID;
+	}
+	// TODO: can we also check whether the packet
+	//
```
dadrian, line 175: ?

src/probe_modules/module_upnp.c
```diff
@@ -176,14 +185,8 @@ void upnp_process_packet(const u_char *packet,
...
 		free(s);
-#if 0
 	} else if (ip_hdr->ip_p == IPPROTO_ICMP) {
-		struct icmp *icmp = (struct icmp *) ((char *) ip_hdr + ip_hdr->ip_hl * 4);
-		struct ip *ip_inner = (struct ip *) ((char *) icmp + ICMP_UNREACH_HEADER_SIZE);
-		// ICMP unreach comes from another server (not the one we sent a probe to);
-		// But we will fix up saddr to be who we sent the probe to, in case you care.
-		fs_modify_string(fs, "saddr", make_ip_str(ip_inner->ip_dst.s_addr), 1);
-		fs_add_string(fs, "classification", (char*) "icmp-unreach", 0);
+		fs_add_constchar(fs, "classification", "icmp-unreach");
```
dadrian, line 189: Are we sure it's an unreach at this point and not another ICMP type? Why isn't this added by `fs_populate_icmp_from_iphdr)`?

src/probe_modules/module_udp.c
```diff
@@ -406,105 +388,75 @@ void udp_process_packet(const u_char *packet, UNUSED uint32_t len,
...
+		size_t ip_inner_len;
+		if (icmp_helper_validate(ip_hdr, len, sizeof(struct udphdr),
+					 &ip_inner,
+					 &ip_inner_len) == PACKET_INVALID) {
+			// find original destination IP and check that we sent a packet
+			// to that IP address
+			uint32_t dest = ip_inner->ip_dst.s_addr;
+			if (!blocklist_is_allowed(dest)) {
+				return PACKET_INVALID;
+			}
```
dadrian, line 457: Isn't this checked by `icmp_helper_validate`?

src/probe_modules/module_upnp.c
```diff
@@ -30,16 +30,17 @@ static const char *upnp_query = "M-SEARCH * HTTP/1.1\r\n"
 
 probe_module_t module_upnp;
 
+static int num_ports;
+
 int upnp_global_initialize(struct state_conf *state)
 {
-	int num_ports = state->source_port_last - state->source_port_first + 1;
-	udp_set_num_ports(num_ports);
+	num_ports = state->source_port_last - state->source_port_first + 1;
```
dadrian, line 37: You need udp_set_num_ports if you're still using `udp_make_packet`

dadrian (approved):

Issue #475, opened 2017-10-04
Packet rate limit option (-r) not respected after a point, sometimes


jhart-r7:
In some testing of zmap with the `-r` and `-B` options, I observed that, at least with my systems, once the effective packet rate reaches ~244,000/s, the packet/s or bandwidth rate limit is ignored and zmap runs unbounded.

We run a UDP scan with something like:

```
/usr/local/sbin/zmap -i eth0 -S <sources> -s <port>  -v1 -b <blacklist_path> -r 350000 --output-filter='success = 1 && repeat = 0 && sport = <port>' -p <port>  -M udp -O csv -f timestamp_ts,saddr,sport,daddr,dport,ipid,ttl,data --probe-args=file:<some_file>
```

TCP studies are run similarly but with the respective TCP options passed instead.  The same issues is observed there.

The 350000 number was presumably chosen at some point in the past as a result of hardware testing.

In testing, however, I've observed that this rate is not respected:

```
 0:00 0%; send: 105 0 p/s (2.84 Kp/s avg); recv: 0 0 p/s (0 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.00%
 0:00 0%; send: 359 347 Kp/s (9.52 Kp/s avg); recv: 0 0 p/s (0 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.00%
 0:01 0%; send: 483036 482 Kp/s (465 Kp/s avg); recv: 95 94 p/s (91 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.02%
 0:02 0%; send: 967710 484 Kp/s (475 Kp/s avg); recv: 198 102 p/s (97 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.02%
 0:03 0%; send: 1452376 484 Kp/s (478 Kp/s avg); recv: 303 104 p/s (99 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.02%
 0:04 0%; send: 1936978 484 Kp/s (480 Kp/s avg); recv: 402 98 p/s (99 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.02%
 0:05 0% (2h06m left); send: 2421597 484 Kp/s (481 Kp/s avg); recv: 502 99 p/s (99 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.02%
[...]

PR #477, opened 2017-10-05
Fix packet rate limiting in send.c under specific conditions. Harden send.c

paul-pearce:
(Note: This branch has a bunch of debug and partial commits in it. Once approved let me squash and edit the commit msg down plz.)

This PR fixes a number of issues:

* `count` and `last_count` were type `uint32_t`. When sending multiple packets (`-P`) this could lead to overflow. They are now `uint64_t`
* We implicitly relied on `count > last_count` and `time > last_time`. These are now assertions (that would have caught the overflow above)
* Our carefully calculated `delay` (used for busy waiting) is thrown away on the first iteration through the loop due to a startup error (`count == 0`). This caused delay to begin at `1` and cause a send spike. This is fixed by not entering the delay adjustment clause on startup (`count == 0`)
* The behavior of our delay calculation code is... problematic. If it ever converges to `delay == 1` and `rate` is set to greater than roughly `1/2 MAX`, where MAX is the max rate the system can send at, we lose control of the loop and emit packets at a rate beyond `rate`, never leaving `delay == 1`. Normally, this shouldn't matter, except when `delay` goes to 1 by mistake (or with the bug above). To harden against this, we now `log_fatal` if `delay -> 1`, both as a precaution to prevent emitting packets faster than `rate` (in the case of a bug or a momentary hiccup) but also as a feedback feature, as if you naturally go to `delay -> 1` it means you've specified a `rate` the hardware can't support.

### DIFF
--- a/src/send.c
+++ b/src/send.c
@@ -271,7 +271,6 @@ int send_run(sock_t st, shard_t *s)
 			last_time = now();
 		}
 	}
-	//printf("XXX Starting: rate: %u, interval: %u, delay: %u\n", zconf.rate, interval, delay);
 	// Get the initial IP to scan.
 	uint32_t current_ip = shard_get_cur_ip(s);
 
--- a/src/send.c
+++ b/src/send.c
@@ -243,8 +243,8 @@ int send_run(sock_t st, shard_t *s)
 	pthread_mutex_unlock(&send_mutex);
 
 	// adaptive timing to hit target rate
-	uint32_t count = 0;
-	uint32_t last_count = count;
+	uint64_t count = 0;
+	uint64_t last_count = count;
 	double last_time = now();
 	uint32_t delay = 0;
 	int interval = 0;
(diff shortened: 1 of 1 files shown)


paul-pearce:
Addresses #475

dadrian:
What happens if you set `-r` close to say, line rate, and hardware supports line rate?

paul-pearce:
@dadrian it should be fine as long as you are not repeatedly under-performing (requesting a rate limit you can't meet) and converging to no delay.

I'm not in a position to test running at such a high rate with our network setup (we'd, um, break stuff). Can you give it a whirl and make sure it's OK?

dadrian:
@paul-pearce just set `-G 00:00:00:00:00:00`

dadrian:
Tested it with near-line-rate speeds on my box. Looks good to me still, although as expected it's less accurate the higher the speed per thread is. Setting `-r 1400000` (1.4Mpps, 0.04 shy of 1gbps), ran at about 1.42Mpps, but adding `-T 2` had a nice and steady 1.4Mpps.

Issue #488, opened 2017-11-09
Incorrect computation of probe packet size results in incorrect send rate when bandwidth (-B) option specified


jhart-r7:
As of late I've been trying to use -B to properly constrain our large scans.  

https://github.com/zmap/zmap/blob/master/src/send.c#L139-L164 looks to be the code that converts the desired bandwidth into the appropriate scan rate given the probe/packet size.

I see a few bugs here:

1) ~~https://github.com/zmap/zmap/blob/master/src/send.c#L144-L146 is adjusting the size of the probes that are less than 84 bytes to exactly 84 bytes for unknown reasons.  My best guess is that this is some sort of word-alignment trick that worked well for TCP but doesn't work with other protocols.  In general this will make for inaccurate calculation of the packet send rate.  When the actual probe is less than 84 bytes, the scan rate will be lower than expected.  This affects TCP syn scans and smaller UDP studies, at least.~~
2) https://github.com/zmap/zmap/blob/master/src/send.c#L140-L142 converts the modules specified/compute packet length from bytes to bits.  The problem is that not every module sets `packet_length` the same way.  For example, UDP probes compute this as ethernet + IP + udp + udp payload (14 + 20 + 8 + N) in one location (https://github.com/zmap/zmap/blob/master/src/probe_modules/module_udp.c#L267-L269), but and just 1 in another (https://github.com/zmap/zmap/blob/master/src/probe_modules/module_udp.c#L877), making the calculations for bandwidth constrained UDP studies possibly inaccurate.  Contrast this with TCP, which correctly sets it to 54 (https://github.com/zmap/zmap/blob/master/src/probe_modules/module_tcp_synscan.c (https://github.com/zmap/zmap/blob/master/src/probe_modules/module_tcp_synscan.c#L169)

jhart-r7:
After letting this sit for some time and recently revisiting it, I'm going to close this and reopen a new issue with more relevant and perhaps actionable bugs.

PR #490, opened 2017-11-10
Clarify/improve rate and bandwidth-based comments and logging

jhart-r7:
In #488 I stumbled with what I thought was a bug.  It isn't.  This PR is the result, which cleans up some comments in this regard and improves the debug output, which I've been using to help understand rate issues.

### DIFF
--- a/src/send.c
+++ b/src/send.c
@@ -135,15 +135,12 @@ iterator_t *send_init(void)
 			    "global initialization for probe module failed.");
 		}
 	}
-	// concert specified bandwidth to packet rate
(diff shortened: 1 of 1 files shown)


zakird:
The reason this is set is because the minimum size of an ethernet frame is 84 bytes: https://en.wikipedia.org/wiki/Ethernet_frame. It doesn't matter if the internal packet is smaller, 84 bytes is still sent on the wire (including the inter-frame gap).

jhart-r7:
Ah yes, that would do it. Thanks for the clarification.

PR #494, opened 2018-01-03
Module to scan for gQUIC support

Eichhoernchen:
This module allows to scan hosts for Google QUIC support.

### DIFF
--- a/src/probe_modules/module_quic_chlo.c
+++ b/src/probe_modules/module_quic_chlo.c
@@ -1,0 +1,428 @@
+/*
+ * Jan Rüth 2018
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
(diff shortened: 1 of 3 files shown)


src/probe_modules/module_quic_chlo.c
```diff
@@ -0,0 +1,428 @@
...
+// start with this
+__uint128_t fnv1a_128(const uint8_t* data, size_t len) {
+	__uint128_t hashHI = 7809847782465536322;
+	__uint128_t hashLO = 7113472399480571277;
+	__uint128_t hash = hashHI << 64 | hashLO;
+	return fnv1a_128_inc(hash, data, len);
+}
+
+void serializeHash(__uint128_t hash, uint8_t out_hash[12]) {
+	// endianess I don't care....
```
justinbastress: If the server is checking this, the endianness is important, no? And if not, then why bother calculating it in the first place?

justinbastress:
This looks like it could be viable, but there isn't bandwidth right now to give it the necessary attention.

PR #498, opened 2018-02-05
Improve rate handling

jhart-r7:
Recently I ran into some trouble with using `zmap` with the rate-based packet limit via `-r`.  In debugging this I didn't see the rate in the logs, which was confusing as compared to what you see with `-B` (for bandwidth).  This patch fixes this such that a proper invocation with `-r` will log the rate.  Also, I've added some simple error handling with rate and bandwidth.

With just rate specified (`-r 1`) correctly:

```
Feb 05 09:37:58.754 [DEBUG] send: rate set to 1 pkt/s
```

With and invalid rate (`-r 0`):

```
Feb 05 09:37:50.202 [FATAL] send: rate impossibly slow
```

With rate and bandwidth (`-r 10 -B 10M`):

```
Feb 05 09:33:20.160 [FATAL] send: Must specify rate or bandwidth, or neither, not both.
```


Explorer1092:
i found error by default.
root@iZbp15nvuk3jmtwsnrjpo3Z:~/zmap-master/src# zmap --bandwidth=10M --target-port=80 --max-targets=10000 --output-file=results.csv
Jul 24 14:54:54.818 [INFO] zmap: output module: csv
Jul 24 14:54:54.818 [FATAL] send: must specify rate or bandwidth, or neither, not both.
root@iZbp15nvuk3jmtwsnrjpo3Z:~/zmap-master/src#

PR #500, opened 2018-02-23
Check output filter expression before evaluating success

### LINKED ISSUE #479: Max results (-N) option not respected exactly
With a few different zmap versions, including https://github.com/zmap/zmap/commit/0258a49568d56a4f62a0c7df741f6a13df1ce0d7 and most recently with the current HEAD of master 5b33c8a9cc94ebc760eb0fe4d037771e4c1e3567, I've noticed that when running with the `-N` option to stop after receiving X results, it doesn't always stop exactly at X.  Sometimes it does, but more often than not it stops at nearly X.

Here is a sample test, scanning a fixed UDP port and telling zmap to stop after 1000 results.  As you can see, it never makes it to 1000:

```
for f in `seq 1 5`; do ./src/zmap -i eth0 -s 123 -b <blacklist> -T 22  --output-filter='success = 1 && repeat = 0 && sport = 123' -p 123  -M udp -O csv -f timestamp_ts,saddr,sport,daddr,dport,ipid,ttl,data -N 1000 --probe-args=file:<probe> -c 30 > foo 2> /dev/null && wc -l foo; done
991 foo
992 foo
987 foo
993 foo
994 foo
```

The only option (that I'm not specifying above) that seems to have any impact on how close to 1000 it gets it the `-r` option for rate limiting.  I suspect this is related to #475 and others.

For context, I only ever use this option for testing a particular zmap configuration.  I don't do this with "full" scans.

ajholland:
This PR fixes #479 

The maximum results field is based upon the number of unique successes. The number of IP's output to the user is not necessarily the number of successes if an output filter is used since the output filter is evaluated after the number of successes is incremented. 

This PR changes the order of operation so the output filter is evaluated first and an IP is only marked as a success if it passes the filter (if one is specified).

### DIFF
--- a/src/recv.c
+++ b/src/recv.c
@@ -69,6 +69,11 @@ void handle_packet(uint32_t buflen, const u_char *bytes)
 	assert(success_index < fs->len);
 	int is_success = fs_get_uint64_by_index(fs, success_index);
 
+	fieldset_t *o = NULL;
+	if (!evaluate_expression(zconf.filter.expression, fs)) {
+		goto cleanup;
+	}
+
 	if (is_success) {
 		zrecv.success_total++;
 		if (!is_repeat) {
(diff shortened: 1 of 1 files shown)


ajholland:
Ah, I see now. Would a better approach be to have a count of IP's that pass the output filter and only halt zmap based on this new counter if an output filter is specified?

zakird:
I suspect so. We may very much well also want to add a counter to the Metadata output for "output results". @dadrian what's your take?

dadrian:
My initial take is `-N=X` should always result in `X` rows of output, but I'm not sure how this should translate into success/failure counts. I think we're probably in "no matter what we do, it will be confusing" territory.

dadrian:
> My initial take is -N=X should always result in X rows of output

Although I realized I already broke this with sharding, where you'll have _at most but probably very nearly X_ rows of output.

zakird:
I agree with David on the functionality of `-N`. I'm leaning towards adding another counter and doing separately at this point.

ajholland:
Makes sense! I'll work on that and PR it, let me know other thoughts if ya got them.

PR #501, opened 2018-02-25
Add formatting enhancements

ajholland:
Went down a rabbit hole when fixing up formatting for my other PR. 

Is there a reason clang-format should be only version 5?

**This PR:**
Changed the format.sh script to support updated versions of clang-format. I also added -style=file since there's a .clang-format file in the directory but it's not currently used in the clang-format command. 

I ran clang-format-5.0 on a VM and it resulted in the same formatting changes as clang-format v 6. (src/send.c, src/probe_modules/module_udp.c). 

I also wrote a script that when run from Travis will error if it detects invalid formatting. 

The reason this isn't currently passing Travis is that clang-format-5.0 isn't in Trusty. And it looks like Travis-ci doesn't currently support Xenial. Could we use the version currently in Trusty (3.6) to test? Otherwise, another workaround is Docker but I'm not sure if that's overkill.


dadrian:
@ajholland IIRC, the older version didn't format the same as the newer version, which is how we ended up with the version check. If Trusty doesn't have the right version, check to see if there's a PPA we can add? I expect Clang has an official one.

ajholland:
@dadrian This works on travis now and is ready for another pass.

src/probe_modules/module_tcp_cisco_backdoor.c
```diff
@@ -44,16 +44,16 @@ static int synscan_init_perthread(void *buf, macaddr_t *src, macaddr_t *gw,
...
+	// instead of settings sequence number to be random for validation
+	// let's instead set to something static so that we can easily
+	// set acknowledgement number. I don't know how integer overflow
+	// is going to act in this.
+	// uint32_t tcp_seq = validation[0];
+	// From Mandiant
+	// 1. To initiate the process, a uniquely crafted TCP SYN packet is sent
+	//	to port 80 of the “implanted” router. It is important to note
+	// that the difference between the sequence and acknowledgment numbers
+	// must be set to 0xC123D. Also the ACK number doesn’t need to be zero.
```
zakird: This indentation doesn't make sense.

ajholland: The entire comment, the part inside the comment, or both? Either way, I agree, I was just making clang-format happy. I'll look into how to fix/why it's doing this.

dadrian: I'd see if you make Clang Format just not edit the interior of comments at all.

ajholland: I updated so it doesn't reflow within comments. I believe it's aligning this block comment with the most recent statement (i.e. line 44 in module_tcp_cisco_backdoor.c) it does the same thing here: https://github.com/zmap/zmap/pull/501/files#diff-092345402e77b3908c6baad58a6851c4R586

zakird:
That part doesn't bother me as much as the fact that it's arbitrarily just indenting blocks that aren't in any function. L47.

ajholland:
Makes sense, I don't know why Clang-format v5 is doing that, and after looking into it I couldn't find any documentation on it. 

I did, however, confirm that clang-format v6 doesn't do this behavior. I could update the build to use clang-format-6.0? There would be no other changes to the style (from updating).

ajholland:
I updated this to use clang-format-6.0 which didn't have the above annoying behavior. 

The main goal of this is to standardize the format used and ensure pull requests follow the format currently established.

dadrian:
How hard is it to install clang-format 6?

dadrian:
Is it brew installable?

>

ajholland:
Yes, `brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/2c2c4b8bfb93e3b409c1c9c51b8a76c3497e0ca0/Formula/clang-format.rb` will install clang-format-6.0. I believe the new default is 7.0.

dadrian:
I made the formatting check soft-fail for now.

PR #503, opened 2018-03-06
add memcached sample probe

justinbastress:
Pads the `stats items` command with the standard UDP packet header.


justinbastress:
Test: `zmap --max-targets=1% --sender-threads=1 -M udp --verbosity=5 --rate=35000 --target-port=11211 --probes=1 --probe-args=file:memcache_11211.pkt --ignore-invalid-hosts`

Issue #478, opened 2017-10-05
Monitor, thread_state, iteration all incorrectly tally IPs instead of packets


paul-pearce:
Well, this is awkward.

All of the code in `monitor.c` for calculating number of packets sent is based on calls to `iterator_get_sent`. This tallies `it->thread_shards[i].state.sent`. This is incredments _per IP_ in `send.c`, not per-packet.  This means when we have multiple packet streams (`-P`) we report IPs/sec, not packets/sec.

For example, at `-r 60000 -P 1` monitor reports 60Kp/s. External monitoring confirms. At `-r 60000 -P 2` monitor reports 30Kp/s, but the line rate is actually 60Kp/s, as `send.c` limits itself correctly.

I think the fix here is to multiply all the monitor statistics by `zconf.packet_streams` but this needs further thought.

Note: The types of the sent variables in `monitor.c` need to be transitioned to `uint64_t` if they begin to accurately reflect the true number of packets sent.

(A cursory glance seems to indicate recv is ok since it's from the pcap, but I didn't check the types.)

dadrian:
This was fixed by #513

PR #513, opened 2018-05-05
Fix Rate and Monitoring 

### LINKED ISSUE #508: -P doesn't respect rate parameter
This is related to #478 

It appears the probes/packet streams operate regardless of the rate specified on the command line. If I specify 
`zmap -p 80 -P 10 -r 1`
I would expect zmap to send one packet every second, however, it sends 10 packets to a single host every second.

This also results in other issues where if I specify a rate of 250,000 pps (and 10 probes), it instead attempts to send packets at 10*250,000 pps and zmap then errors out when attempting to send 2 million pps on the machine I'm working with. 

IMO the expected behavior would be to respect the rate parameter firstly.

I'd be happy to submit a PR to fix this and also work on #478.

ajholland:
This PR Attempts to fix #508 and #478. They're fairly related so I've done them in one PR. 

Using -P for sending multiple packets to an IP currently does not respect the rate functionality. This is an issue because specifying a rate of 250k packets per second and a -P of 10 results in a true send rate of 2.5M packets per second. This PR addresses that by making the packet stream for-loop encapsulate the timing adjustments, sending of the packet, exiting, and monitoring adjustments. The for loop and the goto cleanup statements are the only changes I made to address that. 

You can confirm this change by doing zmap -p 80 -r 1 -P 10 <your choice of IP here> which will send 10 packets to that IP address one second apart. 

The other changes are in regards to the monitor. `state.sent` and `state.tried_sent` were moved into the above mentioned for loop to sum packets sent, not hosts targeted. I've also changed these values to be uint64 type as the number of packets could easily be greater than 2^32. This mostly fixed the monitoring issue but time was still off, multiplying `max_targets` and `max_index` both by packet streams did the trick. I've also changed `send_rate` and `recv_rate` to use ceil since the current monitor outputs those rates as pseudo-integers (floats with 0 decimal digits). 

There are three questions I have regarding this PR:
1. Is Packet Streams the way we want to describe sending multiple packets to a single host? To me, the term packet streams implies sending packets in parallel from multiple NIC's, which is not what is being done. 
2. What should Recv in monitor tally? Currently, it tallies the amount of unique successful packets, which may not be the users chosen output filter. IMO, recv should reflect the output filter and I've made that change in this PR. [Line 168](https://github.com/zmap/zmap/compare/master...ajholland/fixRateAndMonitoring#diff-d32ee771311d9a5d37a54addf7027c09R168)
3. I'm not positive, but there could be other cases where switching the tallying to packets instead of hosts will overflow the uint32 we have, if you want to switch to tallying packets I'd be happy to find these other cases and fix them up in this PR. 

Please let me know if there's any issues or adjustments I can make!


src/monitor.c
```diff
@@ -121,8 +121,10 @@ double compute_remaining_time(double age, uint64_t tried_sent)
 	if (!zsend.complete) {
 		double remaining[] = {INFINITY, INFINITY, INFINITY, INFINITY};
 		if (zsend.max_targets) {
-			double done = (double)tried_sent /
-				      (zsend.max_targets / zconf.total_shards);
+			double done =
+			    (double)tried_sent /
+			    (zsend.max_targets * zconf.packet_streams /
```
dadrian: Will this possible overflow?

src/monitor.c
```diff
@@ -162,7 +165,7 @@ static void export_stats(int_status_t *intrnl, export_status_t *exp,
 	uint32_t total_tried_sent = iterator_get_tried_sent(it);
 	uint32_t total_fail = iterator_get_fail(it);
 	uint32_t total_recv = zrecv.pcap_recv;
-	uint32_t recv_success = zrecv.success_unique;
+	uint32_t recv_success = zrecv.filter_success;
```
zakird: Do we lose something weird here because filter_success might not be unique?

ajholland: I think it's a gain. recv_success is only used in monitor to calculate rate and avg. In my opinion, the user should be presented with a rate and average that matches their chosen output filter, not necessarily unique successes.

zakird: I'm wavering on this. I think we have to decide fundamentally, is filtering _part_ of receiving, or is it _post_ receiving, and we should be consistent with all of the counters.

zakird: along those lines, it would probably be good to go through all the counters in metadata output, updates file, and on-screen updates, and make sure that we have a good definition of all of these and document exactly what they are, and then double check that this matches up with the code base.

ajholland: Can do 👍 I'll look into those counters, see if I can get more comments/consistency. 

FWIW I think that filtering is already a part of receiving since we're filtering success_unique, and therefore not representing the total number of packets ever received. But I can also see the other side too ¯\_(ツ)_/¯

zakird: @ajholland did you ever have a chance to look at those values? I just realized that we never actually merged this PR due to that pending question.

ajholland: So here's all the vars: 
total_sent - total number of **hosts** scanned, this pr makes it _packets_ sent
total_tried_sent - count number of IP's that are iterated through in the cyclic group, including those scanned and not scanned. Used to restrict max targets and compute time remaining. this pr makes this counter also increment per packet sent
recv_success_unique - exported var, set to recv_success which in this PR is filter success 
app_recv_success_unique - alias for zrecv.app_success_unique
total_recv - alias for zrecv.pcap_recv, counter of total number pkts, value set from pcap_stat.ps_recv which is number of packets transited on the network 
percent_complete - dependent on age and remaining_secs
remaining_secs - logic in compute_remaining_time function 
send_rate - total sent / delta
recv_rate - recv success / delta
recv_total_rate - total recv / delta
app_success_rate - app_success / delta where app_success is zrecv.app_success_unique
pcap_drop - number of packets dropped by the driver
time_remaining - alias for remaining_secs 
fail_total - packets failed to reach NIC
hitrate - recv_success * 100 / total_sent 
app_hitrate - app_success * 100 / total_sent

as a note, @zakird I think I'm finding more instances where type overflow would occur because of counting packets instead of ip's (all 5 vars above line 168 here). I'll defer to fixing those until you get a chance to comment on whether that is a change you want. 

When I looked through these, the only ones that seem to be inconsistent in terms of definition and code is sent and tried_sent (and that's fixed in the PR). 

Also, thoughts on renaming zconf.packet_streams?

zakird: Given that this is going to go out with ZMap 3.0, we have the opportunity to change names in this file. I think we should we should do so in order to remove ambiguities. 

Here's what I'm thinking:


`total_sent` - total number of hosts scanned, this pr makes it packets sent *Change: packets_sent. Also add `hosts_scanned`.*
`total_tried_sent` - count number of IP's that are iterated through in the cyclic group, including those scanned and not scanned. Used to restrict max targets and compute time remaining. this pr makes this counter also increment per packet sent *candidate_ips*
`recv_success_unique` - exported var, set to recv_success which in this PR is filter success
app_recv_success_unique - alias for zrecv.app_success_unique *I'm confused about this one is.*
`total_recv` - alias for zrecv.pcap_recv, counter of total number pkts, value set from pcap_stat.ps_recv which is number of packets transited on the network *Change: `response_packets_received`*
percent_complete - dependent on age and remaining_secs
remaining_secs - logic in compute_remaining_time function *-> time_remaining*
send_rate - total sent / delta
recv_rate - recv success / delta *Change: make clear success*
recv_total_rate - total recv / delta *Why does this have a total?*
app_success_rate - app_success / delta where app_success is zrecv.app_success_unique
pcap_drop - number of packets dropped by the driver *Change: pcap_dropped_packets*
time_remaining - alias for remaining_secs *Sure seems like we don't need both of these*
fail_total - packets failed to reach NIC
hitrate - recv_success * 100 / total_sent 
app_hitrate - app_success * 100 / total_sent

zakird: These are all works in progess, not tied to any but wanted to start to get down

zakird: I'm confused about how a bunch of these interact with output filters.

zakird: Note: I'm going to move these to a google spreadsheet and we can discuss there

ajholland: Pinging to see what progress there is on this.

zakird:
Sorry for the delay on this, I can try to map these out on Thursday or
Friday.

zakird:
Alrighty, I've tried to map this out here: https://docs.google.com/spreadsheets/d/1vG4InRqNRF5HmwLXljceFVeZoPL8lHVJYbrTfBvShBE/edit?usp=sharing. @ajholland let me know if this seems sane.

ajholland:
@zakird Looks good to me. I can make those changes in a separate PR? 

The only blocker on this PR is deciding whether the monitor should output response success as filter success. (If so, there's a couple overflows I will fix).

zakird:
I'm cool with a different PR. I don't think it matters too much what we decide here if we're going to immediately replace.

dadrian:
These stats are all tracked in odd places. I'm going to take a stab at moving some of this stuff around, but in the mean time I've updated this PR enough to accurately fix #508 and #478, without breaking time estimation. We now output "packets_sent", and "hosts_scanned". I don't currently output `total_tried_send`, since that information is no longer accessible in the global state object, and we'd need to change where that is tracked for multiple ports support, anyway.

PR #516, opened 2018-05-07
Support non-continuous source IP's

### LINKED ISSUE #338: zmap source IP address option
The `-S` option is very useful to protect scanner itself, right now  `-S` just accept singal IP addr or IP ranges like `10.0.0.0-10.0.0.9`. But sometimes set `-S` to IPs which are far from each other, in different segment, like `10.0.0.1, 10.0.1.15,10.1.1.25`, if the machine allows, I think is much safer than continuous IP ranges.

ajholland:
This PR resolves #338 by implementing the ability to send from non-continuous source IP ranges. 

Ex: `-S 10.0.0.1,10.0.0.3-10.0.0.5,10.0.0.7` will set source IP addresses to one of {10.0.0.1, 10.0.0.3, 10.0.0.4, 10.0.0.5, and 10.0.0.7} for each packet sent but notably not 10.0.0.2 and 10.0.0.6.

I implemented this by parsing the addresses given into a static array of size 256. I don't see a use case where someone would use more than 256 source addresses but this is a design decision I wanted to point out as a limitation that I am open to change. 

I also created a utility.c and utility.h in src/ for better readability. This could all be moved to lib/util.c _except_ for the single function that adds IP addresses to the array. I was unsure of how to best implement that for readability.


zakird:
I'm generally fine with this PR, but I don't think we accurately capture the IPs used for scanning in the metadata file now. If there's a non-contiguous range of IPs, first and last don't have much meaning. We should output the set of IP addresses that's used.

ajholland:
This should not get merged yet. Source IP addresses are reversed. Fix coming soon, and I'll also fix up the metadata output.

ajholland:
@zakird this should be good to go now

PR #524, opened 2018-05-21
Allow ziterate to use up to 65535 shards

jhart-r7:
Following @dadrian's example in #469 

Before:

```
$  ./src/ziterate --shards 12345 -e 54321 --shard 1100 |wc -l
May 21 11:08:35.131 [FATAL] zmap: argument `shard' must be between 0 and 254
```

After:

```
$  ./src/ziterate --shards 12345 -e 54321 --shard 11000 |wc -l
  347911
$  echo $(((2**32)/12345))  
347911
```

### DIFF
--- a/src/ziterate.c
+++ b/src/ziterate.c
@@ -44,8 +44,8 @@ struct zit_conf {
 	int disable_syslog;
 
 	// sharding options
-	uint8_t shard_num;
-	uint8_t total_shards;
+	uint16_t shard_num;
+	uint16_t total_shards;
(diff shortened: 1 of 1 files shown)


zakird:
Seems fine to me?

jhart-r7:
Is there anything I can do to help test this?

zakird:
mostly just harass us to do our jobs :p

dadrian:
@jhart-r7 Sorry for the delay on this one. 😔

PR #529, opened 2018-06-15
Fix bug in prefix length check, typo

### LINKED ISSUE #528: Aborted (core dumped) error when scanning
I just downloaded latest version (commit 0e94a7cd7c3f4fd363af9c39cb882a218439491b) of your code via
```bash
git clone https://github.com/zmap/zmap
```
I built a Docker image using the `Dockerfile` provided, and I was not able to run scan, no matter the options I choose.

I run my container (following Dockerfile instructions) with:

```bash
docker run -it --entrypoint=/bin/bash zmap_ubuntu
```

After that I executed with no success `zmap -p 80` and it gave to me next error:

```bash
zmap: /zmap-master/lib/constraint.c:110: _set_recurse: Assertion `0 <= len && len <= 32' failed.
Aborted (core dumped)
```

ajholland:
I believe this resolves #528 and I confirmed that this works in Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@
 #
 #        docker build -t zmap_ubuntu -f Dockerfile .
 #
-#   * If you wisht to build a specific commit, use the ZMAP_COMMIT build argument.
+#   * If you wish to build a specific commit, use the ZMAP_COMMIT build argument.
 #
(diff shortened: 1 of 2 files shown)


zakird:
Thanks for fixing this, @ajholland

Issue #525, opened 2018-05-22
Unset rate parameter will cause zmap to fatal error


ajholland:
https://github.com/zmap/zmap/commit/3236996d1ff432ee4d067a91d4731b6e84b891d2#diff-7e832c503c89f8b93bb9b0b0d9e8e67dR171

With this line, running `zmap -p 80` quits after stating that the send rate is impossibly slow. (It's really just not set). 

I kinda agree with this change so that the default invocation doesn't send at full-line rate. But I know that also is a feature of zmap? (sending packets at full-line rate). 

Also, this has implications for doing a dry run as I wouldn't expect to have to set a rate parameter to do a dry run.  

At the most, we might have to make a couple changes to make full-line rate the default again. 
At the least, just append the fatal comment to say "rate must be specified". 

@zakird @jhart-r7 @dadrian thoughts on the best approach for this?

jhart-r7:
Yeah, I encountered this same thing just the other day while testing something.  It felt odd to have to set a rate when testing only.  It looks like the handling of line rate (`zconf.rate = 0`) is probably not working at the moment.  There also looks to be a problem if the max bandwidth is specified, because that could result in a `rate` of 0 too.

zakird:
I think it's fine to have a safe default, something like 10K pps and a log message designating this. 0 was supposed to be "full speed, do not rate limit". I think it's clear that the current behavior is broken.

dadrian:
> 0 was supposed to be "full speed, do not rate limit"

We should definitely fix whatever is broken, so long as the concept of "do not rate limit" still exists. This was previously special cased on `zconf.rate = 0`, but it doesn't have to be that way.

zakird:
@ajholland any chance you'd be willing to tackle this? I think we need to specifically:

 (1) support setting `0` as a support rate on the command-line, which does not limit the speed of the scan (and update docs to explain this)

(2) Set default send rate to be something a bit more conservative, maybe 10K pps.

zakird:
This is resolved by #530.

PR #530, opened 2018-06-23
Change Rate Behavior

ajholland:
As a followup to #525 this sets the default rate to 10,000 pps. 

Scan at full line rate by passing `-r 0` 

Behaviors changed: 
-  Default is now 10,000 pps. It was previously full line rate. 
-  Setting 0 as the rate will scan at full line rate.


zakird:
I think it probably makes more sense to have this be a default within ZMap and not just something in the configuration file. I'd like ZMap to act non-broken without a conf.

ajholland:
Fair enough! I'll work on that.

ajholland:
Changed it, wasn't sure if it should have a comment to not be a magic number, but it seemed redundant since it's already the default struct. Wasn't sure if you'd rather have it set as a capitalized const somewhere? maybe just in this file?

zakird:
I think it's probably fine there. We have a bunch of constants just set as defaults there.

src/send.c
```diff
@@ -174,7 +174,7 @@ iterator_t *send_init(void)
 		    zconf.bandwidth, pkt_len / 8, zconf.rate);
 	}
 	// log rate, if explicitly specified
-	if (zconf.rate <= 0) {
+	if (zconf.rate < 0) {
```
zakird: Is rate ever below zero, or was there some weird casting going on where values < 0.5 were being rounded down to zero?

zakird: It looks `zconf.rate` is an integer.

zakird: Looking at L160 above it looks like this should be fine. @ajholland does this also look alright to you?

zakird:
approved. if that comment looks reasonable @ajholland, let's merge it!

src/send.c
ajholland: Lgtm!

ajholland:
Lgtm!

PR #551, opened 2019-03-28
Add option to define payload for icmp_echoscan

sruester:
Probe module icmp_echoscan now supports options to define custom payload
data. Parameter --probe-args can be used with

 * file:/path_to_payload_file
 * text:SomeText
 * hex:5061796c6f6164

When none of them is specified, the probe-module behaves as before, sending
zeroes as echo payload.


dadrian:
@sruester I edited your PR to improve the error handling in `icmp_global_initialize`, and match the default behavior to the behavior prior to the patch (sending 20 bytes of `0x00`.).

@zakird It turns out the ICMP echo probe has been miscalculating lengths, since `sizeof(struct icmp)` is actually _larger_ than the ICMP header. I updated the probe module to calculate length correctly. I also discovered our checksum function didn't correctly handle odd-lengths (since it operates on shorts). This is fine for IP, since IP headers are necessarily a multiple of 4, however ICMP packets can have odd lengths and need to be padded with 0x00 for the checksum. Furthermore, our IP (and ICMP) checksum seems to have had an extra bit shift in it, which I removed. Please review `packet.c`.

I have confirmed all of the ICMP behavior with Wireshark and 8.8.8.8.

PR #555, opened 2019-05-01
move default rate setting later in zmap init

zakird:


jhart-r7:
Confirmed:

```
$  ./src/zmap -p 443 -i en0 -e 1 -T 1  -B 10m  -d -G 00:00:00:00:00:00 -n 2  -v 5 2>&1 | grep rate
May 01 16:08:19.168 [DEBUG] send: using bandwidth 10000000 bits/s for 84 byte probe, rate set to 14880 pkt/s
```

Thank you!  I was working on an identical patch but got distracted ;)

PR #563, opened 2019-08-03
Support for destination port rotation in tcp_synscan

jhalderm:
This patch adds support for destination port rotation to the tcp_synscan probe module.  When the module is invoked with `--probe-args=rotate_dport`, the destination port will be incremented after each probe, wrapping to 0 after 65535.

Combined with ZMap's random target address permutation, this feature can be used to easily sample from the universe of ip:port combinations.  For example, the following command will send 1000 probes to each TCP port 0-65535:

`zmap -p0 -n 65536000 --probe-args=rotate_dport --output-module=csv --output-fields=saddr,sport`

### DIFF
--- a/src/probe_modules/module_tcp_synscan.c
+++ b/src/probe_modules/module_tcp_synscan.c
@@ -8,6 +8,12 @@
 
 // probe module for performing TCP SYN scans
 
+// When executed with "--probe-args=rotate_dport", the module will
+// increment the destination port after transmitting each packet.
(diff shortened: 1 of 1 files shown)


zakird:
@dadrian and I implemented framework-level multi-port scanning this weekend (we use up to a 48-bit group and then choose a dynamic bit boundary between ip and port based on the number of IPs and ports so that we effectively iterate over both), which supersedes this, so I'm going to close.

PR #579, opened 2019-10-24
UDP scan results are unfiltered when multiple zmap processes are running

shellcong:
When multiple zmap processes are doing the UDP scanning on the same host, each zmap process will receive unrelated UDP responses belonging to other zmap processes.

The bug reason is that the zmap UDP scan doesn't check whether the source port of received UDP response is same as the target port of the scanning. This bug only exists when multiple zmap processes run at the same time.


zakird:
This is how ZMap was originally developed. However, many UDP services respond from different ports than they're originally probed on and this cannot be checked globally. If anything, this should be a flag to the UDP module. We do this additional filtering in some UDP modules, e.g., DNS.

PR #587, opened 2020-01-18
Prevent overflow in unlikely seed generation

### LINKED ISSUE #586: IP generation process of ZMap is flawed for some seed values
We have the following bug to report. This bug targets all the known versions of Zmap.

In some (rare) cases, the IP generation of Zmap is flawed and the scan does not cover the full IP range or enter in an infinite loop. This occurs when the number generated by the random generator or the seed specified by the user (with the -e paramater or in the zmap.conf file) is equal to some problematic value. The probability is estimated to 7*10^(-9) (approximatively the euromillion jackpot probability) but is non zero. The bug can be reproduced with the following seed values and the current version of Zmap on github (commit f44fc15):

346296650 -> infinite loop
551370190 -> infinite loop
668513199 -> partial scan (63% of the scan)
950005475 -> partial scan (50% of the scan)
1853945527 -> partial scan (33% of the scan)
2250056444 -> infinite loop
2963139306 -> partial scan (12% of the scan)
8582731091 -> single IP generation
9561776755 -> single IP generation

This list is not exaustive and the number of problematic values is approximatively equal to 2^37.

The command line for testing the bug is the following:
zmap -p 80 -d -e [seed_value] | grep "ip" | wc -l
or equivalently for faster result:
ziterate -b ../conf/blacklist.conf -e [seed_value] | wc -l

The bug is due to the fact that the multiplication in the finite field is coded as a 64-bit unsigned multiplication followed by a modular reduction. This sequence of operations is sound when the two operands are less than 2^32. Some elements of the finite field used by Zmap are however greater than 2^32. In particular, if the primitive element generated by the RNG exceeds 2^32-14, then, during the generation of the cyclic group, one of the multiplication exceeds 2^64 and the multiplication in the finite field is flawed. Moreover, a special case occurs when the primitive element is greater than 2^32. In this case, the function that generates the primitive element (find_primroot) cast the value into 32 bits, resulting in a value that is not a primitive element.

To fix this bug, we recommand to modify the function that generate the primitive element of the finite field. However we do not recommand to change the multiplication of the finite field. Actually using a big number library like mpz will deeply impact the performance of the IP generation process. Moreover the multiplication on 64-bits and the reduction can be coded as two assembly instructions on modern CPUs and it is the most efficient method for coding the generation of the cyclic group. Therefore we only recommand to check that the primitive element is less or equal to 2^32-14. It ensures that all the multiplications with this operand will have a result less than 2^64. This can be done by modifying the function find_primroot in the file cyclic.c:

static uint32_t find_primroot(const cyclic_group_t *group, aesrand_t *aes)
{
  uint64_t retv;
  do {
	  uint32_t candidate =
	      (uint32_t)((aesrand_getword(aes) & 0xFFFFFFFF) % group->prime);
	  if (candidate == 0) {
		  ++candidate;
	  }
	  while (check_coprime(candidate, group) != COPRIME) {
		  ++candidate;
		  // special case where we need to restart check from begin
		  if (candidate >= group->prime) {
			  candidate = 1;
		  }
	  }
	  retv = isomorphism(candidate, group);
  } while (retv > ((1<<32)-14)); //bug fix
	return retv;
}

Rémi Strullu and Johan Mazel

dadrian:
Fixes #586

Co-Authored-By: @rstrullu

### DIFF
--- a/src/cyclic.c
+++ b/src/cyclic.c
@@ -114,7 +114,7 @@ static uint32_t find_primroot(const cyclic_group_t *group, aesrand_t *aes)
 	// The maximum primitive root we can return needs to be small enough such
 	// that there is no overflow when multiplied by any element in the largest
 	// group in ZMap, which currently has p = 2^{32} + 15.
-	const uint64_t max_root = (1 << 32) - 14;
+	const uint64_t max_root = (UINT64_C(1) << 32) - 14;
 
 	// Repeatedly find a generator until we hit one that is small enough. For
 	// the largest group, we have a very low probability of ever executing this
--- a/src/cyclic.c
+++ b/src/cyclic.c
@@ -86,6 +86,9 @@ static cyclic_group_t groups[] = {{// 2^8 + 1
 // Check whether an integer is coprime with (p - 1)
 static int check_coprime(uint64_t check, const cyclic_group_t *group)
 {
+	if (check == 0 || check == 1) {
+		return NOT_COPRIME;
+	}
 	for (unsigned i = 0; i < group->num_prime_factors; i++) {
 		if (group->prime_factors[i] > check &&
 		    !(group->prime_factors[i] % check)) {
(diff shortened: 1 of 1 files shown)


zakird:
This looks good to me. Thanks for the super quick fix @dadrian 

 @rstrullu, if you would be willing to take a second look, that would be much appreciated. @dadrian and I will wait to merge for a couple of days.

rstrullu:
It looks okay to me.

PR #592, opened 2020-03-03
Fix incorrect default prefix length for targets

### LINKED ISSUE #591: Target addresses in whitelist file without prefix are ignored
If target addresses are not specified with a prefix (CIDR), they are ignored by zmap. Therefore even the default examples in the zmap help do not work. zmap uses  an invalid default prefix length of 256, which leads to the whitelist constraint being ignored.

Example:
```
$ sudo ./zmap -p 80 216.58.207.46  -v 5
Mar 03 17:37:13.431 [DEBUG] zmap: zmap main thread started
Mar 03 17:37:13.431 [DEBUG] zmap: syslog support enabled
Mar 03 17:37:13.431 [DEBUG] zmap: requested ouput-module: default
Mar 03 17:37:13.431 [DEBUG] zmap: no output module provided. will use csv.
Mar 03 17:37:13.431 [DEBUG] fieldset: probe module does not supply application success field.
Mar 03 17:37:13.431 [DEBUG] zmap: requested output field (0): saddr
Mar 03 17:37:13.431 [DEBUG] filter: no output filter specified. will use default: exclude duplicates and unssuccessful
Mar 03 17:37:13.431 [DEBUG] constraint: blacklisting 0.0.0.0/0
Mar 03 17:37:13.431 [DEBUG] constraint: Painting value 1
Mar 03 17:37:13.438 [DEBUG] constraint: 0 IPs in radix array, 0 IPs in tree
Mar 03 17:37:13.438 [DEBUG] constraint: 0 addresses (0% of address space) can be scanned
Mar 03 17:37:13.438 [ERROR] blacklist: no addresses are eligible to be scanned in the current configuration. This may be because the blacklist being used by ZMap ((null)) prevents any addresses from receiving probe packets.
Mar 03 17:37:13.438 [FATAL] zmap: unable to initialize blacklist / whitelist
```

ogasser:
The default prefix length if none is given should be 32 and not 256, as
256 is not a valid prefix length.

Resolves #591.

This bug was introduced by 0e94a7cd7c3f4fd363af9c39cb882a218439491b
in PR #526.

### DIFF
--- a/lib/blacklist.c
+++ b/lib/blacklist.c
@@ -104,7 +104,7 @@ void whitelist_prefix(char *ip, int prefix_len)
 
 static int init_from_string(char *ip, int value)
 {
-	int prefix_len = 256;
+	int prefix_len = 32;
 	char *slash = strchr(ip, '/');
 	if (slash) { // split apart network and prefix length
 		*slash = '\0';
@@ -113,7 +113,7 @@ static int init_from_string(char *ip, int value)
 		errno = 0;
 		prefix_len = strtol(len, &end, 10);
 		if (end == len || errno != 0 || prefix_len < 0 ||
-		    prefix_len > 256) {
+		    prefix_len > 32) {
 			log_fatal("constraint",
 				  "'%s' is not a valid prefix length", len);
 			return -1;


zakird:
Thank you!!

PR #610, opened 2020-06-29
batch support

splitice:
Support for sending packets in batches using `--batch` parameter.

No functional change when not used.


zakird:
I kind of like having it exposed. Keeps default behavior the same, but let's people choose something higher at their own preference (though later we may decide to switch to use new socket API that lets you send multiple packets at once to improve performance.

dadrian:
```patch
From bea33f48b5e670c78ee38041a41527e055f07602 Mon Sep 17 00:00:00 2001
From: David Adrian <davidcadrian@gmail.com>
Date: Tue, 30 Jun 2020 16:40:07 -0400
Subject: [PATCH] Correctly handle end-of-shard

---
 src/send.c | 40 ++++++++++++++++++++--------------------
 1 file changed, 20 insertions(+), 20 deletions(-)

diff --git a/src/send.c b/src/send.c
index 03fa617..e780a4f 100644
--- a/src/send.c
+++ b/src/send.c
@@ -340,33 +340,33 @@ int send_run(sock_t st, shard_t *s)
 			}
 		}
 
-		// Check all the ways a send thread could finish and break out
-		// of the send loop.
+		// Check if the program has otherwise completed and break out of the send loop.
 		if (zrecv.complete) {
-			break;
-		}
-		if (s->state.max_targets &&
-		    (s->state.sent >= s->state.max_targets ||
-		     s->state.tried_sent >= s->state.max_targets)) {
-			log_debug(
-			    "send",
-			    "send thread %hhu finished (max targets of %u reached)",
-			    s->thread_id, s->state.max_targets);
-			break;
+			goto cleanup;
 		}
 		if (zconf.max_runtime &&
-		    zconf.max_runtime <= now() - zsend.start) {
-			break;
-		}
-		if (current_ip == ZMAP_SHARD_DONE) {
-			log_debug("send",
-				  "send thread %hhu finished, shard depleted",
-				  s->thread_id);
-			break;
+			zconf.max_runtime <= now() - zsend.start) {
+			goto cleanup;
 		}
 
 		// Actually send a packet.
 		for(int b = 0; b < zconf.batch; b++){
+			// Check if we've finished this shard or thread before sending each
+			// packet, regardless of batch size.
+			if (s->state.max_targets &&
+				(s->state.sent >= s->state.max_targets ||
+				 s->state.tried_sent >= s->state.max_targets)) {
+				log_debug(
+					"send",
+					"send thread %hhu finished (max targets of %u reached)",
+					s->thread_id, s->state.max_targets);
+			}
+			if (current_ip == ZMAP_SHARD_DONE) {
+				log_debug("send",
+					  "send thread %hhu finished, shard depleted",
+					  s->thread_id);
+				goto cleanup;
+			}
 			for (int i = 0; i < zconf.packet_streams; i++) {
 				count++;
 				uint32_t src_ip = get_src_ip(current_ip, i);
-- 
2.21.0 (Apple Git-122.2)
```

splitice:
It looks like the change request has been made, I don't think there is anything you need from me right?

dadrian:
@splitice, you're right, I was able to do the necessary Git incantations to get it on your branch.

dadrian:
It looks like our TravisCI setup is borked, I'll fix that up in a different commit. I'm going to override the status checks and merge.

Thanks for the contribution!

Issue #532, opened 2018-08-01
The --rate and --bandwidth did not work with --list-of-ips-file


jkryanchou:
When I specify  `--list-of-ips-file` to scan an amount of IPs with --rate <rate>, while zmap couldn't send packet in the rate. neither with `--bandwidth`. I wonder if `--list-of-ips-file` could work with the speed parameters (`--rate` or `--bandwidth`) ?

ajholland:
Can you provide the command you ran and verify that you were on the most recent commit of master? 

I don't see a reason --list-of-ips-file would have an effect on --rate and --bandwidth, however we have been tweaking --rate and --bandwidth so a bug in there is not uncommon.

jkryanchou:
Any suggestions on using the `--rate` and `--bandwidth` options with `-I` ?

dadrian:
I thought this was going to be fixed by #513, but it wasn't. I've dug into the code, and figured out what's happening. A few things are interacting:

1. Since `-I` using a paged bitmap, we need to iterate over a cyclic group large enough to cover the full size of the bitmap. This includes the "blocked" addresses that we would have scanned if `-I` wasn't set. This is because we need to check each element of the group against the bitmap to see if it is allowed to be sent. This effectively means we always loop through 4 billion targets, even if only a small number are provided.
2. Debugging the rate limiting, it actually approaches a delay of zero, meaning that the rate limiting code effectively turns itself off for `-I` with small inputs. Nearly all of the send loop is spent simply iterating over the cyclic group and throwing out most IPs are not being on the list.

I tried adjusting our group selection to be based on the number of targets in `-I`. Unfortunately, that results in nearly all of the addresses on a small input not being scanned at all, since with a small group, the chance of actually iterating over something in the bitmap that's allowed is extremely small.

The fundamental issue is that our bitmap is not at all aligned with our cyclic groups. There's probably another bug here as well, where if you have a sufficiently large blocklist, `-I` won't even send to all of the inputs (if we select a group smaller than 2^32).

@zakird I suspect there's some way we could tweak the pbm to be sized the same as the group to fix this problem? Does this make sense to you?

zakird:
I suspect this isn't worth it. If you have a small number of IPs to use, then use whitelist. If you have millions to billions of IPs to scan, then use PBM-based approach. This really only exists for the use-case when you have a huge number of random IPs you need to scan. We could figure out where this threshold is and document it, but I don't think it's worth developing new code for, compared to the other issues pending.

[...]

PR #622, opened 2020-07-14
Be explicit about using large groups with -I

dadrian:
--list-of-ips (-I) only works correctly with groups of order > 2^{32}, since
we have to check each element against a bitmap. This edits the group
selection to always use a large group with using -I.

To make this more obvious to the user, this adds a log_warn if the list
of IPs file is considerably smaller (100000x) than the group size. It
also updates the time remaining calculation to use iterations through
the group when using -I, rather than packets sent.

Address #532 to the extent possible


src/zmap.c
```diff
@@ -826,6 +826,10 @@ int main(int argc, char *argv[])
 	if (!zconf.total_allowed) {
 		log_fatal("zmap", "zero eligible addresses to scan");
 	}
+	if (zconf.list_of_ips_count > 0 && 0xFFFFFFFFU / zconf.list_of_ips_count > 100000) {
+		log_warn("zmap", "list of IPs is small compared to address space, performance will suffer");
+		log_warn("zmap", "consider using an allowlist instead");
```
zakird, line 831: This really feels like it should be a single warning message.

dadrian: It just makes for a very long line

PR #626, opened 2020-10-01
Update README.md

gauravkwatraa:


zakird:
I don't think that we're looking to make this claim.

PR #634, opened 2020-12-18
Remove package managers from README

dadrian:
I think we've been dropped from most of these since we haven't done a stable release in so long, and 2.1.1 depends on old versions of generally unsupported software (including the Redis library)


zakird:
I think this is fine, but we'll want deal with when we release 3.0

zakird (approved):

Issue #636, opened 2020-12-24
PF_RING support not compiling


toomuchio:
Getting compiler errors following the install steps from https://github.com/zmap/zmap/blob/master/10gigE.md and using the latest PF_RING headers.

Bug must have been introduced on https://github.com/zmap/zmap/commit/de9de39de1b371fd5a66a7ff20e5f720a8a3ae5a

/home/test/zmap/src/recv-pfring.c: In function ‘recv_packets’:
/home/test/zmap/src/recv-pfring.c:60:2: error: unknown type name ‘timespec’; use ‘struct’ keyword to refer to the type
  timespec ts = pf_buffer->ts;
  ^~~~~~~~
  struct

jaredthecoder:
I can confirm this is broken for me as well. Same issue, latest PF_RING headers.

jaredthecoder:
Also to note, @toomuchio's fix in their PR worked for me.

zakird:
I believe this has been resolved in #637. Closing.

PR #637, opened 2020-12-24
Fix PF_RING support

toomuchio:
Relates to issue https://github.com/zmap/zmap/issues/636

Might be a cleaner way of passing the timespec back though.


dadrian:
@toomuchio What version of PF_RING are you compiling against?

toomuchio:
The latest one, there was two changes as far as I can see a new flags field when making the PF_RING instance and that struct isn't compatible.

zakird (approved):

PR #638, opened 2021-01-11
Fix ICMP echo scan with `hex:` probe arg prefix not working

marc-etienne:
👋 

The comparaison in case ICMP data is provided with the `hex:` prefix is wrong: the payload is compared with `"hex"` not the full argument. This simple fix makes it work.

Cheers!

### DIFF
--- a/src/probe_modules/module_icmp_echo.c
+++ b/src/probe_modules/module_icmp_echo.c
@@ -79,7 +79,7 @@ int icmp_global_initialize(struct state_conf *conf)
 		icmp_payload_len =
 				fread(icmp_payload, 1, ICMP_MAX_PAYLOAD_LEN, inp);
 		fclose(inp);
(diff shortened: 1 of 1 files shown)


zakird (approved):

zakird:
Thanks for doing this, looks good to me. Unfortunately I can't merge though because the branch is out of date from zmap master (and it doesn't look like I have permissions to merge it in before accepting the PR). Can you update and then I'll merge the PR?

marc-etienne:
I have rebased the commit on `master` and force-pushed, hopefully that'll work.

zakird:
Looks good, will merge once tests pass here in a moment.

zakird:
We still haven't gotten this project off of Travis CI, which of course is rather useless these days. I've checked out and manually run locally.

PR #650, opened 2021-04-29
Add extended validate generator

user318:
Hello.

Some probes may include additional data in probe packets like timestamps, TTL, etc. It may be desirable to include additional information into validation input besides src and dst IPs.
This patch adds additional validation generator function which allows to fill all fields of the aes_input array, so that probe could use it to include additional information to sourch of the validation.
Not sure though about the best naming for the new function.


dadrian (approved):

dadrian:
Thanks!

PR #653, opened 2021-05-11
Implement SYN SCAN with MP_CAPABLE option for multipath TCP support discovery

Ngalstyan4:
[Multipath TCP](http://multipath-tcp.org/) allows changing endpoint IP addresses without breaking existing TCP connections and is becoming more widely deployed. 
TCP SYN scan is a common way to measure support of MPTCP in the wild, both by end hosts and by middleboxes. [Here](http://blog.multipath-tcp.org/blog/html/2018/12/19/which_servers_use_multipath_tcp.html) is a blog post discussing previous such measurement studies. 

This PR implements a probe into zmap that allows to detect multipath-tcp capable end-hosts and interfering middleboxes.
The probe is very similar to tcp_synscan. The main difference is the additional MPTCP TCP option in sent SYN packets and parsing of the reply TCP option in SYNACK packets.

The PR has some rough edges but before polishing these, wanted to see whether it belongs in upstream zmap.

TODOs before merging

 - [ ] module naming
- [ ] Set `th_off` in `make_tcp_header` based on probe needs
- [ ] Standardize or remove `htonll/ntohll` macros
- [ ] ? Avoid copying logic between `synscan`, `synackscan` and `mpscan`

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -29,8 +29,9 @@ set(PROBE_MODULE_SOURCES
     probe_modules/module_icmp_echo.c
     probe_modules/module_icmp_echo_time.c
     probe_modules/module_tcp_synscan.c
+    probe_modules/module_tcp_mpsynscan.c
     probe_modules/module_tcp_synackscan.c
-	#probe_modules/module_tcp_cisco_backdoor.c
+    #probe_modules/module_tcp_cisco_backdoor.c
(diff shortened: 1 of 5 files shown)


dadrian:
The probe module interface has since changed, and the TODO's were not resolved, so I'm closing this out. Feel free to rebase onto main and reopen the PR once it's done.

Issue #631, opened 2020-11-09
redis ip count is different from csv approach 


tahseen2k8:
**Issue**
I am using csv approach which need manual human efforts to get results . tried to use redis to automate but the total count of redis approach is almost double then the count i was getting from csv approach . commands are given below can you guys please help me out in this issue .

**CLI Arguments**
Please paste your ZMap invocation below

```
Redis commands ..
sudo zmap 0.0.0.0/0 -B 300M -p 80 -n 1% --shards=4 --shard=0 --seed=9820 -l zmapLogs/Zmap_B300M_P80_n1%.logs --quiet -f saddr --output-module=redis-csv --output-args=tcp://localhost:6379/queue:2_80
sudo zmap 0.0.0.0/0 -B 300M -p 80 -n 1% --shards=4 --shard=1 --seed=9820 -l zmapLogs/Zmap_B300M_P80_n1%.logs --quiet -f saddr --output-module=redis-csv --output-args=tcp://localhost:6379/queue:2_80
sudo zmap 0.0.0.0/0 -B 300M -p 80 -n 1% --shards=4 --shard=2 --seed=9820 -l zmapLogs/Zmap_B300M_P80_n1%.logs --quiet -f saddr --output-module=redis-csv --output-args=tcp://localhost:6379/queue:2_80
sudo zmap 0.0.0.0/0 -B 300M -p 80 -n 1% --shards=4 --shard=3 --seed=9820 -l zmapLogs/Zmap_B300M_P80_n1%.logs --quiet -f saddr --output-module=redis-csv --output-args=tcp://localhost:6379/queue:2_80
=====================================CSV commands==================
sudo zmap 0.0.0.0/0 -B 300M -p 80 -n 1% --shards=4 --shard=0 --seed=9820 -l zmapLogs/Zmap_B300M_P80_n1%.logs --quiet  -f saddr --output-file=100_percent_p_80.csv

sudo zmap 0.0.0.0/0 -B 300M -p 80 -n 1% --shards=4 --shard=1 --seed=9820 -l zmapLogs/Zmap_B300M_P80_n1%.logs --quiet  -f saddr --output-file=100_percent_p_80.csv
sudo zmap 0.0.0.0/0 -B 300M -p 80 -n 1% --shards=4 --shard=2 --seed=9820 -l zmapLogs/Zmap_B300M_P80_n1%.logs --quiet  -f saddr --output-file=100_percent_p_80.csv
[...]

Issue #564, opened 2019-08-08
Redis output ignores default output filter --output-filter="success = 1 && repeat = 0"


eplewis89:
**(see my last comment, this is the path that I took discovering this issue)**

During a zmap scan, I'm writing results to redis as resdis-csv format, and my redis queue is gaining a considerable amount more results than what is being reported in the zmap output.

For example, the 36.0.0.0/8 subnet at a 200/second rate is showing only 23 responses (after 10 minutes), but my redis queue has a factor of 100x the responses in the subnet.

I'm not sure if this is a bug with Redis, or some weird threading issue, but I'm running a second scan along side of the redis scan, outputting results to a file. The second scan is gathering results at a much slower rate even though the rate limit is set to 200.

What info do you need from me to look into this?

eplewis89:
More info, two scans are currently running, port 2223 on the 36.0.0.0/8 CIDR

First scan (the "redis" scan), the command is: "36.0.0.0/8 -p 2223 --output-module=redis-csv --output-args=tcp://x.x.x.x:6379/testresults:2223 -r 200 -f saddr,saddr-raw,sport,timestamp-str"

Second scan (the "file" scan), the command is: "36.0.0.0/8 -p 2223 -o /zmap-results/2223/36.0.0.0-subnet-8.txt -r 200 -f saddr,daddr,sport,dport,seqnum,acknum,timestamp-str"

In the first scan, at 1 minute and 13k packets sent, the redis queue has 1000 items in it, but the zmap output says "1:04 0% (23h left); send: 12671 197 p/s (197 p/s avg); recv: 5 0 p/s (0 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.04%"

In the second scan, at the 1 minute mark, the zmap output says "1:04 0% (23h left); send: 12803 199 p/s (199 p/s avg); recv: 2 0 p/s (0 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.02%" and the file has 2 results in it.

This is weird right?

eplewis89:
Ok I think I found the issue. In the scan that outputs to file, inherently, the output filter is set to the following: --output-filter="success = 1 && repeat = 0"

Somehow the redis module is ignoring this flag and writing IP addresses out (I didn't look to see whether they were duplicates or not).

dadrian:
[...]

PR #668, opened 2021-08-01
Sane default behavior

zakird:
Our default behavior was a bit confusing. This changes the logic to be that default behavior is only if none of {output-filter, output-fields, output-module} are defined. Otherwise, ZMap leaves "default mode" and operates based on those fields default empty values. There are now debug messages that explain this explicitly. This should help prevent the type of confusion in #564, #631, and #655.

### DIFF
--- a/src/state.h
+++ b/src/state.h
@@ -105,7 +105,7 @@ struct state_conf {
 	char **destination_cidrs;
 	int destination_cidrs_len;
 	const char *raw_output_fields;
-	char **output_fields;
+	const char **output_fields;
 	struct output_filter filter;
 	char *output_filter_str;
 	struct fieldset_conf fsconf;
--- a/src/state.h
+++ b/src/state.h
@@ -104,7 +104,7 @@ struct state_conf {
 	char *custom_metadata_str;
 	char **destination_cidrs;
 	int destination_cidrs_len;
-	char *raw_output_fields;
+	const char *raw_output_fields;
 	char **output_fields;
 	struct output_filter filter;
 	char *output_filter_str;
@@ -126,6 +126,8 @@ struct state_conf {
 	int max_sendto_failures;
 	float min_hitrate;
 	int data_link_size;
+	int default_mode;
+	int no_header_row;
 #ifdef PFRING
 	struct {
 		pfring_zc_cluster *cluster;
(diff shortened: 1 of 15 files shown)


src/state.h
```diff
@@ -126,6 +126,8 @@ struct state_conf {
 	int max_sendto_failures;
 	float min_hitrate;
 	int data_link_size;
+	int default_mode;
+	int no_header_row;
```
dadrian, line 130: Should we get rid of `filter_duplicates` and `filter_unsuccessful` from `zconf` as well? They can't be set on the CLI, and I think they always match `default_mode`.

src/zmap.1.ronn
```diff
@@ -52,8 +52,8 @@ on a gigabit network connection, reaching ~98% theoretical line speed.
 
    * `-r`, `--rate=pps`:
      Set the send rate in packets/sec. Note: when combined with --probes,  this is
-     total packets per second, not IPs per second. Setting the rate to 0 will scan 
-     at full line rate. Default: 10000 pps. 
+     total packets per second, not IPs per second. Setting the rate to 0 will scan
+     at full line rate. Default: 10000 pps.
```
dadrian, line 56: Should we default to 0?

src/zmap.c
```diff
@@ -388,11 +389,25 @@ int main(int argc, char *argv[])
...
+	// addresses that responded successfully. We only use this simple "default"
+	// mode if none of {output module, output filter, output fields} are set.
+	zconf.default_mode = (!(args.output_module_given || args.output_filter_given || args.output_fields_given));
+	if (zconf.default_mode) {
+		log_info("zmap", "By default, ZMap will output the unique IP addresses "
+				"of hosts that respond successfully (e.g., SYN-ACK packet). This "
+				"is equivalent to running ZMap with the following flags: "
+				"--output-module=csv --output-fields=saddr --output-filter='"
+				"success=1 && repeat=0' --no-header-row. "
+				"If you want all responses, select a different output module or "
```
dadrian, line 402: Not so much select a different output module, as "explicitly set an output module", since the output module is still CSV here. I don't think you can set `output-module=default` anymore?

dadrian (approved):
LGTM % nits

PR #671, opened 2021-08-01
fieldsets govern that the right fields are added at the right times

zakird:
Good: This finds problems
Bad: DNS module is borked and now doesn't run because it's not safe


zakird:
Going to close and merge as part of ICMP handling code.

PR #681, opened 2021-11-10
Remove `ntp_make_packet` since it is dead code

jonasbb:
The function is not called anywhere. The module's make_packet uses `udp_make_packet`.
The type of `ntp_make_packet` is also not compatible with `probe_make_packet_cb` any longer. For example the `size_t *buf_len` argument is missing.


dadrian (approved):

dadrian:
Thanks, I meant to do this when I cleaned up the probe module interface.

PR #684, opened 2022-02-19
Probe-modules: IPIP

yannayl:
Add a probe module to scan for hosts with IPIP protocol (4) supported.

These hosts may be used for IP spoofing, NAT traversal or just lengthening the path which an IP traverses along the network.

Usage example:
`zmap  --target-port=53  --probe-module=ipip -i eth0.2`

For uses see: [PoC||GTFO 0x21:3](https://www.alchemistowl.org/pocorgtfo/pocorgtfo21.pdf)
This issue was [responsibly disclosed](https://www.kb.cert.org/vuls/id/636397) two years ago to Cert CC


zakird (approved):

zakird:
@phillip-stephens Looked at this briefly and the probe module itself looks reasonable. Would you be willing to help fix up types/tests issues, and merge?

phillip-stephens:
Validated that after fixing up merge conflicts from the past 2 years of changes, the scanning behavior was not changed.
Specifically, ran `sudo ./src/zmap -p 53 --probe-module=ipip -i enp0s8 -B 100M -c 2 1.1.1.1/13` on commit `0285d19` and `eab1103` and the hit hosts were identical (3 found hosts). The scan did _not_ return anything when run locally on my MacBook, but when run on an Ubuntu VM in the lab, it did. I'm chocking this up to something strange with my Mac's networking (valid?).


Let me just validate the packets generated look fine in Wireshark and then I'm good to merge.

Issue #702, opened 2022-10-13
Segmentation fault (core dumped) 


ShenBanigans:
# Describe the bug

ZMAP3 will periodically, and randomly segfault. The command-line options used do not seem to mater. **This would also happen with ZMAP2 which is why I have started trying out ZMAP3 but it still happens in version 3**. It is extremely difficult for me to provide information to reproduce the issue due the way in which I am utilizing your tool for research but I will do my best: 

I have a built a python script for scanning large numbers of subnets for specific ports, it organizes the folder structure the way I want and reads a list of subnets in CIDR format from a specified text file and loops through a list of ports I want to scan and then uses python subprocess module to execute zmap against the subnet for the port I want. Say I have 2700 subnets in the input file, it will sometimes segfault after getting through a couple of hundred, sometimes it cant even get through 10 invocations, sometimes it will run for hours with no problems. It is an extremely bug since I cannot automate my research as I have to constantly check on if zmap has crashed or not when invoked by python.

I have used zmap for decades and have only started noticing this bug during a research task that involves scanning thousands of subnets. **Please note that it does not matter what command line arguments I am using with ZMAP and it also does not matter what subnets are being scanned** - For example say I have a list of 2500 CIDR ranges and I am looping through the list and say for example it segfaults on the 10th one in the list, if i were to re-run my script which loops through the list and invokes zmap, it wont segfault again at the 10th one in the list, it might way past that before it happens again, it might segfault on the first one in the list, it might ssegfault on the 200th one etc.

# CLI Arguments
I am invoking zmap using a python script with subprocess.check_call() and the command line arguments I use vary depending on what I am wanting to do with a specific subnet. However here is an example of a segfault situation where it crashed out when scanning the 4th subnet out of a list of 1024 total. **Please note the bug happens when scanning TCP or UDP, on any port, for any subnet size, the below is just an example of one time it happened when I was scanning for UDP/161.** 

## Example One

```
zmap -M udp --probe-args=file:/home/user/Downloads/zmap-3.0.0-beta1/examples/udp-probes/snmp1_161.pkt -i enp0s31f6 -T 1 -B 1M -p 161 --output-filter="success = 1" -O csv 41.72.64.0/19 -o 41.72.64.0_19-161
[...]

PR #705, opened 2022-10-30
log_fatal on bad send

dadrian:
Addresses #702, maybe.


dadrian:
Previously, if `send` failed to initialize, we'd still try to run the regular finalization / summary / etc. Presumably, data structures that relies are on not fully set up if the failure in send was early enough.

dadrian (commented):
add missing semicolon

zakird (approved):

PR #707, opened 2023-01-10
Use CLOCK_MONOTONIC for rate limiting (fix #706)

### LINKED ISSUE #706: ZMap crashes when the host clock goes backwards
**Describe the bug**

ZMap randomly crashes with an assertion failure:

```
3:22 16% (17m left); send: 7052636 35.0 Kp/s (34.9 Kp/s avg); recv: 733668 3.39 Kp/s (3.63 Kp/s avg); drops: 0 p/s (0 p/s avg); hitrate: 10.40%
3:23 17% (17m left); send: 7087436 34.8 Kp/s (34.9 Kp/s avg); recv: 736653 2.98 Kp/s (3.63 Kp/s avg); drops: 0 p/s (0 p/s avg); hitrate: 10.39%
zmap: /root/zmap/src/send.c:322: send_run: Assertion `t > last_time' failed.
```

**CLI Arguments**
```
zmap -r 35000 -M icmp_echoscan -w zmap_targets.csv -b /root/blacklist_v4 -o icmp.csv
```

**Expected behavior**

ZMap should not crash.

**Environment:**
 - OS: Ubuntu 16.04.7
 - Version: https://github.com/zmap/zmap/commit/7471d5413c5453c81

**Additional context**

There is probably an issue on our host causing the clock to go backwards (e.g. NTP).

However, I noticed that ZMap uses `gettimeofday` which [makes no guarantee](https://man7.org/linux/man-pages/man2/settimeofday.2.html) on the monotonicity of the clock:

> The time returned by gettimeofday() is affected by discontinuous jumps in the system time (e.g., if the system administrator manually changes the system time).  If you need a monotonically increasing clock, see [clock_gettime(2)](https://man7.org/linux/man-pages/man2/clock_gettime.2.html).

Here is a small program to reproduce the issue (on affected hosts):

```c
#include <stdio.h>
#include <sys/time.h>

// From zmap/src/logger.c
double now(void) {
    struct timeval now;
    gettimeofday(&now, NULL);
    return (double) now.tv_sec + (double) now.tv_usec / 1000000.;
}

int main() {
    double last_time = now();
    while (1) {
        double t = now();
        if (t < last_time) {
            printf("Non-monotonic clock: now=%f last=%f diff=%f\n", t, last_time, t - last_time);
            break;
        }
        last_time = t;
    }
    return 0;
}
```

```bash
gcc -O3 main.c && ./a.out
# After a few seconds:
# Non-monotonic clock: now=1671705231.484864 last=1671705232.725731 diff=-1.240867
```

This issue is bound to happen randomly when ZMap runs for a long time.
Would it makes sense for ZMap to use `clock_gettime` instead of `gettimeofday`?

maxmouchet:
This replaces calls to `gettimeofday` with `clock_gettime(CLOCK_MONOTONIC, ...)` in the rate limiting code.
This fixes #706 where ZMap can crash if the host clock goes backwards.

Feel free to suggest a better name/location for the `steady_now` function!

### DIFF
--- a/lib/logger.c
+++ b/lib/logger.c
@@ -267,6 +267,19 @@ double now(void)
 	return (double)now.tv_sec + (double)now.tv_usec / 1000000.;
 }
 
+double steady_now(void)
+{
+#if defined(_POSIX_TIMERS) && defined(_POSIX_MONOTONIC_CLOCK)
+	struct timespec tp;
+	clock_gettime(CLOCK_MONOTONIC, &tp);
+	return (double)tp.tv_sec + (double)tp.tv_nsec / 1000000000.;
+#else
+	struct timeval now;
+	gettimeofday(&now, NULL);
+	return (double)now.tv_sec + (double)now.tv_usec / 1000000.;
+#endif
+}
+
 size_t dstrftime(char *buf, size_t maxsize, const char *format, double tm)
 {
 	struct timeval tv;
--- a/lib/logger.h
+++ b/lib/logger.h
@@ -60,7 +60,18 @@ void check_and_log_file_error(FILE *file, const char *name);
 
 size_t dstrftime(char *, size_t, const char *, double);
 
+// The number of seconds and microseconds since the Epoch.
 double now();
+
+// The number of seconds and nanoseconds since an unspecified point in time.
+// On supported hosts, this value is guaranteed to never decrease.
+//
+// According to the POSIX specification the `clock_gettime` function is part
+// of the Timers option and may not be available on all implementations.
+//
+// On hosts where a monotonic clock is not available, this falls back
+// to `gettimeofday` which was ZMap's original implementation.
+double steady_now();
 
 #ifdef __cplusplus
 }
--- a/src/send.c
+++ b/src/send.c
@@ -73,15 +73,6 @@ void sig_handler_decrease_speed(UNUSED int signal)
 		 zconf.rate);
 }
 
-// A monotonic clock whose value is guaranteed to never decrease.
-// This is useful for rate limiting.
-double steady_now(void)
-{
-	struct timespec tp;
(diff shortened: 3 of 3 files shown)


zakird:
Overall this looks good, but reading the POSIX standard, it does seem like we do need to ensure support for `CLOCK_MONOTONIC` before we use it . Only `CLOCK_REALTIME` is guaranteed to exist on POSIX. (I would imagine that most systems we use support this, but we've been bitten a bunch of times in the past — people run ZMap on strange platforms.) If I'm reading the docs right, we can check whether `_POSIX_TIMERS` is defined to ensure `clock_gettime` is available. If that looks good, we also need to check that `_POSIX_MONOTONIC_CLOCK` also is also there.

> On POSIX systems on which these functions are available, the symbol _POSIX_TIMERS is defined in <[unistd.h](https://linux.die.net/include/unistd.h)> to a value greater than 0. The symbols _POSIX_MONOTONIC_CLOCK, _POSIX_CPUTIME, _POSIX_THREAD_CPUTIME indicate that CLOCK_MONOTONIC, CLOCK_PROCESS_CPUTIME_ID, CLOCK_THREAD_CPUTIME_ID are available

maxmouchet:
I've moved the function to `logger.c` and added a bit of documentation in `logger.h`.
I've also added the POSIX check, which seems to work as expected on my machine.
I thought about adding a `#pragma message(...)` directive to log which implementation is chosen but I don't know if it's necessary and/or portable across compilers.

zakird:
I moved the functions to `util.c` since they don't really belong in `logger.c`. Not your bad — I don't know why `now()` was defined there int the first place. I merged into main in #708.

maxmouchet:
Awesome, thanks!

Issue #562, opened 2019-07-10
icmp_echo_time timestamps not accurate


Maximilian-Fuchs:
I'm trying to measure round trip times (RTTs) with the icmp_echo_time module.
In the results I can find between 15% and 45% negative round trip times, which shouldn't be possible. The positive RTTs in my results also seem way to high.

There seems to be an issue with the timestamps that are written at send and receive of an ICMP packet.

I tried slow measurements (at 100kbit/s) but that didn't help.
Also tried measurements from different machines with the same faulty results.

Has anyone managed to get valid measurements with the icmp_echo_time probe module yet?

Maximilian-Fuchs:
Ok, it was the fault of our evaluation script 🤦‍♂️.
But the absolute RTTs are still higher than the ones optained with `ping` command from the bash.

Maybe we can get a more accurate timestamp if we use libpcap for obtaining the receive timestamp instead of gettimeofday.

Greetings,
Max

rwlrwlrwlrwl:
Will you plan to fix this issue?

PR #726, opened 2023-08-18
Store link-layer timestamp in icmp_echo_time module

maxmouchet:
Fixes inaccurate RTTs when using the `icmp_echo_time` module (https://github.com/zmap/zmap/issues/562).

This follows up on https://github.com/zmap/zmap/pull/607 which passes the pcap/pf_ring timestamp to the packet handler. The information was there but was seemingly never used.

This adds two fields to the output of the `icmp_echo_time` module based on the capture timestamp: `recv_timestamp_ts` and `recv_timestamp_ns`.

Test:

```bash
./src/zmap -M icmp_echo_time -O json 8.8.8.8 \
  -f sent_timestamp_ts,sent_timestamp_us,timestamp_ts,timestamp_us,recv_timestamp_ts,recv_timestamp_ns > res.json
```

RTT with original timestamp:

```bash
jq '((.timestamp_ts * 1e6 + .timestamp_us) - (.sent_timestamp_ts * 1e6 + .sent_timestamp_us)) / 1e3' res.json
1044.48
```

RTT with link-layer timestamp:

```bash
jq '((.recv_timestamp_ts * 1e6 + .recv_timestamp_ns / 1e3) - (.sent_timestamp_ts * 1e6 + .sent_timestamp_us)) / 1e3' res.json
17.838
```

RTT with ping:

```bash
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=116 time=20.2 ms
64 bytes from 8.8.8.8: icmp_seq=2 ttl=116 time=18.3 ms
64 bytes from 8.8.8.8: icmp_seq=3 ttl=116 time=16.3 ms
```

### DIFF
--- a/src/probe_modules/module_icmp_echo_time.c
+++ b/src/probe_modules/module_icmp_echo_time.c
@@ -178,8 +178,18 @@ static void icmp_echo_process_packet(const u_char *packet,
 
 	struct icmp_payload_for_rtt *payload =
 	    (struct icmp_payload_for_rtt *)(((char *)icmp_hdr) + 8);
-	fs_add_uint64(fs, "sent_timestamp_ts", (uint64_t)payload->sent_tv_sec);
-	fs_add_uint64(fs, "sent_timestamp_us", (uint64_t)payload->sent_tv_usec);
+
+	uint64_t sent_timestamp_ts = (uint64_t)payload->sent_tv_sec;
+	uint64_t sent_timestamp_us = (uint64_t)payload->sent_tv_usec;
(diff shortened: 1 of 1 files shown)


maxmouchet:
I've made a few tweaks to my original commit:
- Store `recv_timestamp_us` instead of `recv_timestamp_ns` for consistency
  - Send timestamp is only available at the microsecond resolution so it's not very useful to store the receive timestamp at the nanosecond resolution.
- Add `rtt_us` so that the user doesn't have to compute the RTT himself

zakird (approved):

PR #733, opened 2023-10-13
Use interface with default route on Linux

Tim---:
It seems that patch bf5ed24c1535e0342631cf9ad904fa59720bc46b changed the way zmap finds the default interface.

It used to use `get_default_gw` to find the interface with the default route. Now it uses `pcap_lookupdev` which takes the first interface suitable for packet capture, but does not ensure that it has the default route.

This patch reverts to the old behavior and is consistent with `get_gateway-bsd.h`.

### DIFF
--- a/src/get_gateway-linux.h
+++ b/src/get_gateway-linux.h
@@ -21,21 +21,6 @@
 
 #define GW_BUFFER_SIZE 64000
 
-char *get_default_iface(void)
-{
-	char errbuf[PCAP_ERRBUF_SIZE];
-	char *iface = pcap_lookupdev(errbuf);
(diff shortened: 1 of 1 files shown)


phillip-stephens (changes requested):
Overall, this looks good.
The one issue I see is that when I ran `Valgrind` against `main` without and with this code change, with had an extra 16 bytes still reachable.

We have over 4 MB of total memory `still reachable` according to Valgrind in `main`, so this is relatively a small issue (that this [issue](https://github.com/zmap/zmap/issues/743) was raised to address), but we shouldn't introduce more `still reachable` memory, IMO.

PR #740, opened 2023-11-14
Update manual pages to add details to --probes to contrast with --retries

phillip-stephens:
- added details to `--probes` to contrast with `--retries`
- re-generated manual pages with `make manpages`
- added `.DS_Store` to `.gitignore`. Looks like it can be auto-generated on MacOS, don't want that slipping into the repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ CMakeCache.txt
 install_manifest.txt
 _CPack_Packages/*
 src/ztests
+*.DS_Store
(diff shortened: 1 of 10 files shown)


zakird:
This broadly looks good to me. We should update here to say that probes isn't number of packets per _IP_ but per target _IP/port_ pair. (This is my fault, but we should update it).

Can you make sure that we're also updating the `--help` text as well? Unfortunately the two are maintained separately. I think the text there is generally much shorter, but I did notice that under `probes` definition there it says packets per IP, which should be packets per ip/port too.

phillip-stephens:
@zakird I addressed your concerns and updated the wiki with the `multi-port` usages. If you could re-review whenever you have a chance, that'd be great!

zakird (approved):

PR #751, opened 2023-12-08
Add dynamic packet batching to take advantage of `sendmmsg` on Linux

phillip-stephens:
- [x] Updated manpages
- [x] Updated `--help` text

## Testing
Tested on both Mac and Ubuntu

### Correctness
Ran against a smaller subnet with both `main` and `Phillip/batch-packets` and both returned same output

### Performance
Inputs
- `sender-threads=4`
- `batch=64`
12% speedup compared to `main` on lab Ubuntu box (1.40 Mp/s vs. 1.25 Mp/s)

Inputs
- `sender-threads=1`
- `batch=64`
48% speedup compared to `main` on lab Ubuntu box (860 Kp/s vs. 551 Kp/s)


src/CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ set(SOURCES
     ports.c
     recv.c
     send.c
+#    send-linux.c
```
zakird, line 50: It looks like there's some cleanup needed here.

src/send-bsd.h
```diff
@@ -13,6 +13,7 @@
 #include <sys/time.h>
 #include <sys/ioctl.h>
 #include <fcntl.h>
+#include "./send.h"
```
zakird, line 16: Looks like should be down a line

src/send-linux.c
```diff
@@ -0,0 +1,93 @@
+//
```
zakird, line 1: In general, we don't have authorship on flies. Instead, this needs to have the ZMap copyright blurb on the top like the other source files.

src/send-bsd.h
```diff
@@ -34,4 +35,48 @@ int send_packet(sock_t sock, void *buf, int len, UNUSED uint32_t idx)
...
 }
 
+// Since BSD doesn't have the sendmmsg syscall leveraged in send-linux.c, this just wraps the single send_packet call.
+// However, the behavior in sendmmsg is to send as many packets as possible until one fails, and then return the number of sent packets.
+// Following the same pattern for consistency
+// Returns - number of packets sent
+// Returns last error code if no packets could be sent successfully
+int send_batch(sock_t sock, batch_t* batch, int retries) {
+	int packets_sent = 0;
+	int retry_ct = 0;
```
zakird, line 45: This looks like it can be defined in the loop, it's not used outside of it.

src/CMakeLists.txt
phillip-stephens: Good point, I've removed this line and moved it to only being added if compiling on a non-Apple, non-BSD system.

src/send-bsd.h
phillip-stephens: Addressed!

src/send-linux.c
phillip-stephens: Addressed!

src/send-bsd.h
zakird: The compiler is realistically going to take care of this — it's not really worth worrying about.

src/send-linux.c
```diff
@@ -0,0 +1,105 @@
...
+	int rv = 0;
+	for (int i = 0; i < retries; i++) {
+		// according to manpages
+		// On success, sendmmsg() returns the number of messages sent from msgvec; if this is less than vlen, the
+		//       caller can retry with a further sendmmsg() call to send the remaining messages.
+		// On error, -1 is returned, and errno is set to indicate the error.
+		rv = sendmmsg(sock.sock, &msgvec, batch->len, 0);
+		if (rv < 0) {
+			// only retry if all messages failed to send
+			perror("error in sendmmsg");
```
zakird, line 99: This should be a `log_fatal`

src/send-linux.c
```diff
@@ -0,0 +1,105 @@
...
+		} else {
+			break;
```
zakird, line 101: I'm not sure if I understand the logic here offhand. If there's a major error we retry, but if there's only the ability to do a partial send, then the function returns. The comment below (when the function is called) says:

> // whether batch succeeds or fails, this was the only attempt. Any re-tries are handled within batch

but I don't think this is true because if a fraction of messages aren't sent, then they aren't retried.

src/send-pfring.h
```diff
@@ -16,9 +16,9 @@
 #error "Don't include send-bsd.h or send-linux.h with send-pfring.h"
 #endif
 
-int send_run_init(sock_t socket)
+int send_run_init(anonymous struct socket, int i, struct shard *pShard)
```
zakird, line 19: The anonymous struct here worries me a little bit versus having two explicitly different functions that encapsulate the different behavior.

phillip-stephens: I'm honestly not sure why I made this change. I didn't touch anything related to `pfring`, and all the other `send_run_init`'s have the old format. So I'll revert this change.

src/send-linux.c
phillip-stephens: That's true, and if this is the error-handling we want I should correct the comment. 
From my reading of the `man sendmmsg`, the caller _could_ attempt to re-send the packets starting with the first that failed, but I figured this wasn't really necessary. The user will get an error message on why the batch failed and can retry.

Do you think it'd be better to:

- Batch succeeds -> return
- Batch partially fails, -> retry the failed packets unless you're out of retries, then return
- Batch completely fails -> retry the entire batch, unless you're out of retries, then return

?

zakird: If we didn't have retry behavior and were to just crash on send fail, it'd be different, but I think in our case, we're making the promise to the user _to retry_ based on having the command line option. I think that we should retry for the packets that did not successfully send. Generally, we shouldn't resend packets that were sent successfully since this creates load on the destination system.

src/send-linux.c
phillip-stephens: Addressed!

zakird:
@phillip-stephens it looks like this is now conflicting with another PR that we merged. Can you resolve conflicts?

phillip-stephens:
@zakird LMK if you think differently, but I'm thinking that we wait on this PR and I'll integrate it after I finalize the `liburing` work. `liburing` will be used by most linux users on newer distros anyway. Either way, I've pushed my retry logic updates to address your fair point [here](https://github.com/zmap/zmap/pull/751#discussion_r1424496153).

src/send-linux.c
```diff
@@ -0,0 +1,115 @@
...
+		total_packets_sent += rv;
+		if (rv == num_of_packets_in_batch){
+			// all packets in batch were sent successfully
+			break;
+		}
+		// batch send was only partially successful, we'll retry if we have retries available
+		log_warn("batch send", "only successfully sent %d packets out of a batch of %d packets", total_packets_sent, batch->len);
+		// remove successfully sent packets from batch for retry
+		current_msg_vec = &msgvec[total_packets_sent];
+		num_of_packets_in_batch = batch->len - total_packets_sent;
```
dadrian, line 112: This looks like the it assumes the failed packets are always the back of `msgvec`...is that accurate? Could an arbitrary subset fail?

If it's always the back of the vector that failed, then I would recommend adding a comment to clarify.

src/send-linux.h
```diff
@@ -5,12 +5,20 @@
...
  * of the License at http://www.apache.org/licenses/LICENSE-2.0
  */
-
 #ifndef ZMAP_SEND_LINUX_H
 #define ZMAP_SEND_LINUX_H
 
-#include "../lib/includes.h"
+#include <stdlib.h>
+#include <stdio.h>
+#include <netinet/ip.h>
```
dadrian, line 13: FWIW, I think we try to put the Linux headers after the C stdlib, but we're probably not consistent about it.

C stdlib -> Linux/System Headers -> ZMap headers

src/send.c
```diff
@@ -489,3 +482,25 @@ int send_run(sock_t st, shard_t *s)
 	log_debug("send", "thread %hu cleanly finished", s->thread_id);
 	return EXIT_SUCCESS;
 }
+
+batch_t* create_packet_batch(uint8_t capacity) {
+	batch_t* batch = malloc(sizeof(batch_t));
```
dadrian, line 487: If you want to be real cool, allocate this all in a single malloc call and slice off chunks of it youself, such that all of the memory for batch and what it points to is contiguous. It should help with cache locality.

dadrian (approved):

zakird:
@phillip-stephens I would err towards getting this merged now since the work is complete. I don't see a ton of reason to hang around as a PR. Can you tag me once you've had a chance to address @dadrian's comments?

src/send-linux.c
phillip-stephens: Yep, that was _my_ interpretation of the [man pages](https://man7.org/linux/man-pages/man2/sendmmsg.2.html),
```
If an error occurs after at least one message has been sent, the
       call succeeds, and returns the number of messages sent.  The
       error code is lost.  The caller can retry the transmission,
       starting at the first failed message, but there is no guarantee
       that, if an error is returned, it will be the same as the one
       that was lost on the previous call.
```

So  I don't believe an arbitrary subset could fail. I'll add a clarifying comment to call this out!

src/send.c
phillip-stephens: Oooo, that's a great idea!

phillip-stephens:
Addressed @dadrian's suggestions (thanks for that neat cache locality idea, I'm going to have to use that in the future). Assigning back to you, @zakird

src/send-bsd.h
```diff
@@ -34,4 +35,48 @@ int send_packet(sock_t sock, void *buf, int len, UNUSED uint32_t idx)
...
+			const char *addr_str =
+			    inet_ntop(
+				AF_INET, &addr,
+				addr_str_buf,
+				INET_ADDRSTRLEN);
+			if (addr_str != NULL) {
+				log_debug(
+				    "send",
+				    "send_packet failed for %s. %s",
+				    addr_str,
```
zakird, line 70: Can you fix up indentation here — pretty sure `errno` stuff isn't pushing us past 80char world

src/send-bsd.h
```diff
@@ -34,4 +35,48 @@ int send_packet(sock_t sock, void *buf, int len, UNUSED uint32_t idx)
...
+// Since BSD doesn't have the sendmmsg syscall leveraged in send-linux.c, this just wraps the single send_packet call.
+// However, the behavior in sendmmsg is to send as many packets as possible until one fails, and then return the number of sent packets.
+// Following the same pattern for consistency
+// Returns - number of packets sent
+// Returns last error code if no packets could be sent successfully
+int send_batch(sock_t sock, batch_t* batch, int retries) {
+	int packets_sent = 0;
+	int rc = 0;
+	for (int i=0;i<batch->len;i++) {
+		int retry_ct = 0;
```
zakird, line 47: This looks like it can be defined in the loop.

zakird, line 47: This can be defined on line 48, no need to define outside of loop since only used within loop.

src/send-bsd.h
```diff
@@ -34,4 +35,48 @@ int send_packet(sock_t sock, void *buf, int len, UNUSED uint32_t idx)
...
 
+// Since BSD doesn't have the sendmmsg syscall leveraged in send-linux.c, this just wraps the single send_packet call.
+// However, the behavior in sendmmsg is to send as many packets as possible until one fails, and then return the number of sent packets.
+// Following the same pattern for consistency
+// Returns - number of packets sent
+// Returns last error code if no packets could be sent successfully
+int send_batch(sock_t sock, batch_t* batch, int retries) {
+	int packets_sent = 0;
+	int rc = 0;
+	for (int i=0;i<batch->len;i++) {
```
zakird, line 46: Can you space to match what's on L48 style wise?

src/send-bsd.h
```diff
@@ -34,4 +35,48 @@ int send_packet(sock_t sock, void *buf, int len, UNUSED uint32_t idx)
...
+				log_debug(
+				    "send",
+				    "send_packet failed for %s. %s",
+				    addr_str,
+				    strerror(
+					errno));
+			}
+		}
+	}
+	if (packets_sent == 0) {
```
zakird, line 76: I don't think that I follow why there needs to be different logic here for `packets_sent == 0`

zakird (approved):

src/send-bsd.h
phillip-stephens: Should have added a comment explaining the special handling to mimic linux's `send_mmsg` call. I've updated this!

PR #752, opened 2023-12-09
Set default senders as min(4, number of cores on host)

phillip-stephens:
This should give us a more sane default, especially on hosts with less cores.

- [x] update manages
- [x] updated `--help`

# Testing

Ran on Mac, compiles and runs correctly

To test if it picks up cores correctly, ran on a Ubuntu VM with 2 cores, it detected this correctly.

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -44,6 +44,14 @@ int max_int(int a, int b)
 	return b;
 }
 
+int min_int(int a, int b)
+{
+	if (a >= b) {
+		return b;
+	}
+	return a;
+}
+
 void enforce_range(const char *name, int v, int min, int max)
 {
 	if (check_range(v, min, max) == EXIT_FAILURE) {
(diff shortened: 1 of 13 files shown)


zakird:
I think that we want _max_, not _min_ here. Having more senders than cores is going to cause crazy contention, and for the most part, you don't need more than four threads for performance. I would also probably do `cores - 1` so that you have a core for processing received packets.

zakird:
Sorry, I was totally backwards in that last comment (deleted). Should have coffee before looking at it. That said, I think we should let people set a higher number of cores if they want to. It might make sense to have something like default = -1, and if that's teh case, then we set to `min(4, number of cores on host)`, otherwise we follow what was provided on the commend line.

zakird (approved):

zakird:
@phillip-stephens I have pushed slightly updated logic. Can you take a quick look, and if it looks good to you too, we can merge?

phillip-stephens:
@zakird LGTM, thanks! Merging now

PR #753, opened 2023-12-12
phillip/748: resolves #748 where using the `-I` CLI flag would cause an assertion failure whenever it was run

### LINKED ISSUE #748: Latest ZMap seems to break with the `-I` flag
> Notice: Please don't file a bug report for questions on how to use ZMap, etc. Check out our [Resources](https://github.com/zmap/zmap/tree/main#resources) section for FAQ and feel free to chime in on [Github Discussions](https://github.com/zmap/zmap/discussions) if you're still unclear and we'll be happy to assist!

**Describe the bug**
Using the `-I` flag seems to always lead to a fatal error:
```
zmap: /home/pstephens/zmap/lib/constraint.c:253: constraint_lookup_index: Assertion `index < con->root->count' failed.
Aborted
```

**CLI Arguments**
Please paste your ZMap invocation below

```
sudo zmap -p 80 -I large_ip_set.txt -o "output.csv" -B 500M
sudo zmap -p 80 -I small_set_ips.txt -o "output.csv" -B 500M
```

**Example Target IP**

See attached IP list files
[large_ip_set.txt](https://github.com/zmap/zmap/files/13517752/large_ip_set.txt)
[small_set_ips.txt](https://github.com/zmap/zmap/files/13517754/small_set_ips.txt)

**Expected behavior**
I'd expect the command to scan the given IP file without erroring, unless there's a formatting issue with the text file.

**Environment:**
 - OS: Ubuntu 18.04 and MacOS when compiled from source
 - Version: Github HEAD

**Additional context**
Full output of using command
```
pstephens@scratch-04-a:~$ cat small_set_ips.txt
57.34.30.22
107.237.157.122
94.231.62.61
102.159.87.118
236.130.238.137
84.231.23.216
209.161.244.186
213.62.11.23
230.0.255.180
40.230.12.94
pstephens@scratch-04-a:~$ sudo zmap -p 80 -I small_set_ips.txt -o "output.csv" -B 500M
Nov 30 21:14:39.825 [INFO] zmap: By default, ZMap will output the unique IP addresses of hosts that respond successfully (e.g., SYN-ACK packet). This is equivalent to running ZMap with the following flags: --output-module=csv --output-fields=saddr --output-filter='success=1 && repeat=0' --no-header-row. If you want all responses, explicitly set an output module or set --output-filter="".
Nov 30 21:14:39.825 [INFO] dedup: Response deduplication method is full
Nov 30 21:14:39.856 [WARN] zmap: list of IPs is small compared to address space. Performance will suffer, consider using an allowlist instead
Nov 30 21:14:39.901 [INFO] recv: duplicate responses will be excluded from output
Nov 30 21:14:39.901 [INFO] recv: unsuccessful responses will be excluded from output
 0:00 0%; send: 0 0 p/s (0 p/s avg); recv: 0 0 p/s (0 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.00%
zmap: /home/pstephens/zmap/lib/constraint.c:253: constraint_lookup_index: Assertion `index < con->root->count' failed.
Aborted
```

phillip-stephens:
Took me longer than it should have to find.

## Bug Description
This bug was introduced in commit: [d9c0](https://github.com/zmap/zmap/commit/d9c0efaab18a7251ec18a46cdc7c8c0f4c1f43c9)

Prior to this commit, when the user wanted to use a file for a list of IP's (`-I`), that'd trigger this code in[ iterator.c.](https://github.com/zmap/zmap/commit/d9c0efaab18a7251ec18a46cdc7c8c0f4c1f43c9#diff-dd379cbd28d1d4637f823178215f790099fb7d312e5d09a244890c6b2897bd11L65-L68)

```
-       uint64_t num_addrs = blocklist_count_allowed();
-       uint64_t group_min_size = num_addrs;
-       if (zconf.list_of_ips_filename) {
-               log_debug("send", "forcing max group size for compatibility with -I");
-               group_min_size = 0xFFFFFFFF;
-       }
```

This commit moved/changed this logic to [send.c](https://github.com/zmap/zmap/commit/d9c0efaab18a7251ec18a46cdc7c8c0f4c1f43c9#diff-33c4faff4dd3631a679cba7327bb63d614db66b2d3d29899ecb9bf30462a9085R91-R94).
```
+       uint64_t num_addrs = blocklist_count_allowed();
+       if (zconf.list_of_ips_filename) {
+               log_debug("send", "forcing max group size for compatibility with -I");
+               num_addrs = 0xFFFFFFFF;
+       }
+       it = iterator_init(zconf.senders, zconf.shard_num,
+                       zconf.total_shards, num_addrs, zconf.ports->port_count);
```
The issue is that by setting `num_addrs=0xFFFFFFFF`, we allow the iterator to generate values 0-2^32. However, our radix tree only has values for the `blocklist_count_allowed`, or all the IP addresses that are NOT black/blocklisted (looks like ~3.7B values based on `Dec 12 14:53:23.952 [DEBUG] constraint: 3702243328 IPs in radix array, 15104 IPs in tree.`). Therefore our iterator can generate values that would be outside our radix tree (>3.7B), and therefore violate the assertion [here](https://github.com/zmap/zmap/blob/main/lib/constraint.c#L253).

## Fix
I think we can just remove the setting of the `group_min_size` and let `num_addrs = blocklist_count_allowed()` as it is in the base case.
The old logic of setting `group_min_size = 0xFFFFFFFF` for `zconf.list_of_ips_filename` is retained [here] (https://github.com/zmap/zmap/blob/main/src/iterator.c#L79) for all cases, including `-I` since even if we're only scanning 1 port, the value is always >= 2^32. IMO removing this will solve the bug and not have any repercussions.


zakird:
This seems reasonable to me to allow the blacklist to set the number of scanned IPs because it already has to be scanning `0/0` for this to be working in the first place and the blacklist should know this like in any other case. @dadrian can you confirm this logically makes sense.

@phillip-stephens can you make sure that there's a safety check that if you're using `-I` then the user is scanning `0/0`?

phillip-stephens:
> This seems reasonable to me to allow the blacklist to set the number of scanned IPs because it already has to be scanning `0/0` for this to be working in the first place and the blacklist should know this like in any other case. @dadrian can you confirm this logically makes sense.
> 
> @phillip-stephens can you make sure that there's a safety check that if you're using `-I` then the user is scanning `0/0`? I'm a little bit worried that there could be weird behavior here where we don't actually ensure that the the whole cyclic group / blacklist is being iterated over and checked here.

Hmm, I'm not sure what you mean. The `-I` option is used when the user wants to scan a specific set of IP's, given in a file. Ex: `-I small-set-of-ips.txt`. So what would this safety check do in this case?

zakird (approved):

PR #755, opened 2023-12-14
Align `--retries` default with `--help` and manual pages

phillip-stephens:
- aligned the `--help` **[text](https://github.com/zmap/zmap/blob/main/src/zmap.1.ronn?plain=1#L89)** and true default for `--retries`.
- alphabetized the state file fields to make it easier to find a field


zakird (approved):

zakird:
These should obviously match. I wonder though whether we want to change the default behavior, or if we want to have the help text match the behavior.

phillip-stephens:
Yeah, I thought about that. Figured I'd issue the PR and we can discuss here. My impression is that packets failing to send would be a relatively rare event and making the default `--retries` = 0 would be acceptable. The user should get a message informing them of any send failures and they can adjust if required.

If you don't have any objections, I can modify the PR to do this.

PR #759, opened 2023-12-27
Gracefully handle IPv6 addresses in `blocklist.conf`

### LINKED ISSUE #756: Handle IPv6 addresses/subnets in `blocklist.conf` gracefully
Currently if you run a ZMap scan with a `blocklist.conf` that includes a IPv6 IP or subnet, it errors out.

## IPv6 IP
IPv6 IP in `blocklist.conf` - `2001:db8:3333:4444:5555:6666:7777:8888`

Output
```shell
~/zmap$ sudo ./src/zmap -p 53 -r 128 -N 10 -b /etc/zmap/blocklist.conf

Dec 15 17:20:56.204 [INFO] zmap: By default, ZMap will output the unique IP addresses of hosts that respond successfully (e.g., SYN-ACK packet). This is equivalent to running ZMap with the following flags: --output-module=csv --output-fields=saddr --output-filter='success=1 && repeat=0' --no-header-row. If you want all responses, explicitly set an output module or set --output-filter="".
Dec 15 17:20:56.204 [INFO] dedup: Response deduplication method is full
Dec 15 17:20:56.204 [ERROR] constraint: '2001:db8:3333:4444:5555:6666:7777:8888' is not a valid IP address or hostname
Dec 15 17:20:56.204 [FATAL] blocklist: unable to parse blocklist file: /etc/zmap/blocklist.conf
```

## IPv6 Subnet
IPv6 Subnet in `blocklist.conf` - `2001:db8:3333:4444:5555:6666:7777:8888/64`

Output
```shell
sudo ./src/zmap -p 53 -r 128 -N 10 -b /etc/zmap/blocklist.conf

Dec 15 17:20:24.559 [INFO] zmap: By default, ZMap will output the unique IP addresses of hosts that respond successfully (e.g., SYN-ACK packet). This is equivalent to running ZMap with the following flags: --output-module=csv --output-fields=saddr --output-filter='success=1 && repeat=0' --no-header-row. If you want all responses, explicitly set an output module or set --output-filter="".
Dec 15 17:20:24.559 [INFO] dedup: Response deduplication method is full
Dec 15 17:20:24.559 [FATAL] constraint: '64' is not a valid prefix length
```

phillip-stephens:
Fixes #756 

## Changes
- added a function to check if a string was an IPv6 address or subnet

## Testing
Tested that IPv4 addresses or subnets didn't trigger a `log_fatal`

Tested with the following IPv6 tests
With subnet mask
```
pstephens@scratch-04-a:~/zmap-ubuntu-dev$ sudo ./src/zmap -p 53 -r 128 -N 10 -b /etc/zmap/blocklist.conf
....
Dec 27 17:50:24.508 [FATAL] constraint: 2001:0db8:85a3::/48 IPv6 subnets are not supported
```
With compressed IPv6 address
```
pstephens@scratch-04-a:~/zmap-ubuntu-dev$ sudo ./src/zmap -p 53 -r 128 -N 10 -b /etc/zmap/blocklist.conf
...
Dec 27 17:50:38.421 [FATAL] constraint: 2001:0db8:85a3:: IPv6 subnets are not supported
```
With uncompressed IPv6 address
```
pstephens@scratch-04-a:~/zmap-ubuntu-dev$ sudo ./src/zmap -p 53 -r 128 -N 10 -b /etc/zmap/blocklist.conf
...
Dec 27 17:51:00.807 [FATAL] constraint: 2001:0db8:85a3:0000:0000:8a2e:0370:7334 IPv6 subnets are not supported
```


lib/blocklist.c
```diff
@@ -6,22 +6,24 @@
...
-#include "blocklist.h"
 
 #include <errno.h>
 #include <stdio.h>
 #include <assert.h>
 #include <stdint.h>
 #include <stdlib.h>
+#include <stdbool.h>
 #include <string.h>
+#include "xalloc.h"
```
zakird, line 17: I think that this move isn't correct since this is ZMap code, not a stdlib include.

lib/blocklist.c
```diff
@@ -102,8 +104,33 @@ void allowlist_prefix(char *ip, int prefix_len)
 	_add_constraint(addr, prefix_len, ADDR_ALLOWED);
 }
 
+static int is_ip_ipv6(char *ip) {
+	// don't modify the input string
+	int length = strlen(ip);
+	char *new_str = (char*)malloc(length);
```
zakird, line 110: This should use `xmalloc`, not `malloc`. This also makes the following check unnecessary.

lib/blocklist.c
```diff
@@ -102,8 +104,33 @@ void allowlist_prefix(char *ip, int prefix_len)
...
+	if (new_str == NULL) {
+		log_fatal("blocklist", "could not allocate memory for IPv6 validation");
+	}
+	strncpy(new_str, ip, length);
```
zakird, line 114: There's an off by one here. `The strlen() function returns the number of characters that precede the terminating NUL character.`. However, `new_str` is defined as the length of the string without the null byte at the end. Last, `strncpy` is pretty much deprecated and we should use `strlcpy`.

zakird: Last, I'll just note that `strndup` is a function and I suspect that's what you're looking for :)

lib/blocklist.c
```diff
@@ -102,8 +104,33 @@ void allowlist_prefix(char *ip, int prefix_len)
...
+		// set maskChar char to NULL char so we can check if subnet is valid IPv6
+		*maskChar = '\0';
+	}
+	// attempt conversion of IP into IPv6 struct to check if IP is an IPv6 address
+	struct in6_addr ipv6_addr;
+	if (inet_pton(AF_INET6, new_str, &ipv6_addr) == 1) {
+		return true;
+	}
+	return false;
+}
```
zakird, line 127: Given that we called `malloc` in the function, we should cleanup the memory allocation here.

lib/blocklist.c
```diff
@@ -102,8 +104,33 @@ void allowlist_prefix(char *ip, int prefix_len)
...
+
 static int init_from_string(char *ip, int value)
 {
+	if (is_ip_ipv6(ip)) {
+		log_fatal("constraint", "%s IPv6 subnets are not supported", ip);
```
zakird, line 132: I think that we want to ignore IPv6 subnets, not fatal on them. The idea is that many blacklists will have both IPv4 and IPv6 entries in them. Since ZMap is an IPv4-only tool right now, we simply want to ignore the IPv6 subnets.

lib/blocklist.c
phillip-stephens: Ahh whoops, assumed that was where `malloc`/`calloc`/etc. came from. I'll move it back!

lib/blocklist.c
phillip-stephens: Ahh yikes, yep good catch.

lib/blocklist.c
```diff
@@ -102,8 +102,33 @@ void allowlist_prefix(char *ip, int prefix_len)
 	_add_constraint(addr, prefix_len, ADDR_ALLOWED);
 }
 
+static int is_ip_ipv6(char *ip) {
+	// get length of string including the null-byte
+	int length = strlen(ip) + 1;
+	// don't modify the input string
+	char *new_str = strndup(ip, length);
```
zakird, line 109: I don't really understand the intent behind calling `strlen` and then passing that into `strndup`. This feels equivalent to using `strdup`. If you want to have a limited length, we should use `strndup` and check that the length is below a certain size using `strnlen`. Otherwise, we might as well just use `strdup`.

lib/blocklist.c
```diff
@@ -102,8 +102,33 @@ void allowlist_prefix(char *ip, int prefix_len)
...
+	// check if there's a subnet maskChar
+	char *maskChar = strchr(new_str, '/');
```
zakird, line 111: Can you use one variable naming type in this function? We have both `new_str` and `maskChar`? :) I believe that the ZMap style guide says to use `mask_char`.

zakird (changes requested):

lib/blocklist.c
phillip-stephens: 😬 yikes yep, ofc

lib/blocklist.c
phillip-stephens: Yeah, not sure what I was thinking, `strdup` looks ideal for this.

phillip-stephens:
@zakird I've addressed your comments, re-tested on the `scratch VM` and it's looking good. Re-assigned back to you.

zakird (approved):

PR #764, opened 2024-01-19
Add option to change "recursion desired" bit in DNS probe module

### LINKED ISSUE #763: DNS probe module: let user set recursion desired bit
Hello, I would like to do DNS probes that have the dns_rd bit set to 0. Currently, it is always set to 1 (see https://github.com/zmap/zmap/blob/main/src/probe_modules/module_dns.c#L190).

I have make a fork of the project in which I've changed this value, however, this means that I have to recompile each time I want to change the value of this bit. Instead, I had ther idea to make this configurable to the user and i wanted to ask you if that would be a desirable change to zmap in your opinion. I'd offer to you to implement this change, however, I am not yet sure on how exactly to realize it. For example, it is conceivable to change the syntax to the value of `--probe-args`. Currently. users can specify questions like `--probe-args="A,google.com;A,example.org"`. I can imagine allowing e.g. `--probe-args="A,google.com;A,example.org;rd"` to explicitly desire recursion and e.g. `--probe-args="A,google.com;A,example.org;nrd"` to explicitly set this bit to 0. 

Do you think that this would be a worthwhile change to the project? Or do you see this change breaking backwards-compatibility? I'd be thankful for any feedback.

gutjuri:
The DNS probe module now *additionally* supports options strings like `A:rn,google.com`. The `rn` has the effect that the "recursion desired" bit will be unset in the respective DNS query packet. If `:rn` is omitted, the behaviour is not changed (compared to before this update). Therefore, the change maintains backward-compatibility.

Examples of supported behaviours:
- `--probe-args="A,google.com"`: rd-bit set to 1
- `--probe-args="A:rn,google.com"`: rd-bit set to 0
- `--probe-args="A,google.com;A:rn,example.org"`: rd-bit in the query for google.com set to 1, and in the query to example.org set to 0.

I have tested these behaviours using the above examples with wireshark.

closes #763 

# ToDo

- [ ] Update Documentation (where?)


zakird:
Thanks for putting this together. In terms of help text, it would be best to update the blurb in the probe module registration, which is where we store documentation specific to a given probe module.

This blurb: 
```
    .helptext =
	"This module sends out DNS queries and parses basic responses. "
	"By default, the module will perform an A record lookup for "
	"google.com. You can specify other queries using the --probe-args "
	"argument in the form: 'type,query', e.g. 'A,google.com'. The module "
	"supports sending the the following types: of queries: A, NS, CNAME, SOA, "
	"PTR, MX, TXT, AAAA, RRSIG, and ALL. The module will accept and attempt "
	"to parse all DNS responses. There is currently support for parsing out "
	"full data from A, NS, CNAME, MX, TXT, and AAAA. Any other types will be "
	"output in raw form."
```

zakird:
@phillip-stephens This looks good to me overall. Once we have documentation included in the PR, can you get merged?

gutjuri:
I have updated the documentation and marked the PR as ready to review. Please let me know of there is anything I can do to improve the proposed code changes. Thank you for your time!

phillip-stephens (approved):
LGTM, thanks @gutjuri !!!

src/probe_modules/module_dns.c
```diff
@@ -81,14 +81,18 @@ static int num_ports;
 
 const char default_domain[] = "www.google.com";
 const uint16_t default_qtype = DNS_QTYPE_A;
+const uint8_t default_rdbit = 0xFF;
```
phillip-stephens, line 84: Shouldn't this be `0x01` since the default value is 1, not 255? Or does any non-zero value have the same effect when processed by the probed target.

phillip-stephens: Update: just noticed that this is used to set a bit-value, so this is fine. Approving

phillip-stephens (approved):
LGTM, thanks again

PR #767, opened 2024-01-23
Add optional CLI parameter for Linux `enable-liburing` to use the async I/O library for increased send performance

phillip-stephens:
Scratch Space as I clean up this PR
## Todo

- [ ] Write up manual page
- [ ] Update --help to say `enable-liburing` is only available with select distros + versions + document what those versions are
- [ ] Rewrite/Remove `10gigE.md` file
- [ ] Update wiki/INSTALL docs to mention how you should _not_ use `--enable-liburing` with un-supported distros/versions
- [ ] Document what hit-rates I achieved
- [x] Add conditional compilation so the code will compile successfully on Linux/non-Linux and older linux distros without liburing and those with
- [x] Add CLI flag to enable the `liburing` code flows

## Context
`liburing` is the API library that wraps the underlying `io_uring` Linux async I/O library. It provides an elegant interface to submit I/O tasks asynchronously to the kernel, vastly improving networking performance. 

Various linux distros ship with various linux kernels, each of which has a different version of `io_uring` baked in. Additionally, the package managers of different distro versions will ship different versions of `liburing`. Adding on to the confusion, the developers of `liburing`/`io_uring` have changed, updated, and added features rapidly over the past couple years. This means a guide written 2 years ago may very well be out-of-date. A small example is how this otherwise great [guide](https://unixism.net/loti/tutorial/sq_poll.html) to `liburing` mentions how to use submission queue polling, you _must_ register the file handlers before. This isn't true in recent versions.

Takeaway here is I'll need to test what versions have the features our code uses and list these as `Supported Liburing/io_uring Versions`


zakird:
Given lack of improvement over `send_mmsg`, closing.

PR #770, opened 2024-02-02
Prevent --batch from overflowing the uint8_t it is being stored in

droe:
Prevents crash for `--batch` sizes of multiples of 256, which stored in `zconf.batch` of type `uint8_t` end up as 0.

/cc @phillip-stephens


phillip-stephens (approved):
LGTM, thanks for this!

PR #772, opened 2024-02-02
Multiple fixes for BSD get_gateway code, remove libdnet

droe:
Fixes and modernisations for BSD `get_gateway` code:

- Fix `get_iface_hw_addr()` on macOS, where `--source-mac` was needed in at least some cases, depending on libdnet, by rewriting it using getifaddrs(3) in lieu of libdnet.  getifaddrs(3) is generally portable across all BSD variants and it had already been used for `get_iface_ip`.
- Fix `get_hw_addr()` on FreeBSD, where `--gateway-mac` was needed, by rewriting it using `sysctl(2)` in lieu of libdnet.
- While here, fix two file descriptor leaks and one memory leak in `get_default_iface` and `get_default_gw`.
- Remove the libdnet dependency; it had already been removed for Linux, now it is no longer needed for macOS / BSD either.

Tested on macOS Sonoma as well as FreeBSD 14.0.

(Apologies for extending the scope of an already open PR, but it seemed to make more sense to reason about these changes as a whole, not separately.  Happy to split into multiple PRs if you prefer.)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -97,4 +97,3 @@ option. For example, to install it in `$HOME/opt` run
     make -j4
     make install
     ```
-- On some BSD systems, you man need to manually add `-D_SYSTYPE_BSD` to your CFLAGS.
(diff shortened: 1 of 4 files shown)


phillip-stephens:
This looks good, definitely going to be nice to not have to supply `--source-mac` on MacOS. Thanks @droe!!! I've checked out the branch and sanity tested it, seems good. Merging.

phillip-stephens (approved):
LGTM, thanks @droe !

PR #776, opened 2024-02-10
Support IP layer mode on macOS

droe:
- Implement IP layer mode on macOS using `socket(PF_INET, SOCK_RAW, IPPROTO_RAW)` and `setsockopt(IP_HDRINCL)`
- Teach recv-pcap to handle BSD loopback encapsulation as used by macOS `utun` interfaces
- While here fix file descriptor leaks in bpf failure paths

Deliberately ifdef'ed macOS-specific code with `__APPLE__` such that in the future, the code can be shared amongst macOS and BSDs lacking sendmmsg.  If/when I get around to implement sendmmsg for BSD, I'll likely propose to split send code into bsd-legacy and bsd-sendmmsg, instead of send-bsd and send-mac with largely duplicate code for the "no sendmmsg available" case.

### DIFF
--- a/src/recv-pcap.c
+++ b/src/recv-pcap.c
@@ -69,6 +69,11 @@ void recv_init(void)
 			  errbuf);
 	}
 	switch (pcap_datalink(pc)) {
+	case DLT_NULL:
+		// utun on macOS
+		log_debug("recv", "BSD loopback encapsulation");
+		zconf.data_link_size = 4;
+		break;
 	case DLT_EN10MB:
 		log_debug("recv", "Data link layer Ethernet");
 		zconf.data_link_size = sizeof(struct ether_header);
@@ -84,7 +89,7 @@ void recv_init(void)
 		break;
 #endif
 	default:
-		log_error("recv", "unknown data link layer");
+		log_error("recv", "unknown data link layer: %u", pcap_datalink(pc));
 	}
 
 	struct bpf_program bpf;
--- a/src/send-mac.h
+++ b/src/send-mac.h
@@ -39,7 +39,7 @@ int send_packet(sock_t sock, void *buf, int len, UNUSED uint32_t retry_ct)
 		struct ip *iph = (struct ip *)buf;
 #ifdef __APPLE__
 		// The len and off fields need to be in host byte order on macOS.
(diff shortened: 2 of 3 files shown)


zakird:
Overall makes sense. One question:

> Deliberately ifdef'ed macOS-specific code with __APPLE__ such that in the future, the code can be shared amongst macOS and BSDs lacking sendmmsg.

This makes sense, but I'm not sure if the code in that block is actually Apple-specific. It seems that the host-order part:

		 BSD loopback encapsulation; the link layer header is a	4-byte
		 field,	 in  host  byte	 order,	 containing  a	PF_ value from
		 socket.h for the network-layer	protocol of the	packet

Is for BSD more broadly.

droe:
The ifdef `__APPLE__` is on the send path, the byte order comment there applies only to IP header fields when sending through a raw IP socket (as opposed to BPF device, used by zmap when `-X` is not given), and is independent of the link layer type of the interface.  The packet sent to the socket begins with the IP header and has no link layer header.  macOS requires the off and len fields of the IP header in host byte order when sending via raw IP sockets.  FreeBSD used to require them in host byte order too, back when Apple forked the BSD portion of XNU from FreeBSD 4, but I think since around FreeBSD 11, FreeBSD takes those fields in network byte order.  When sending with those fields in host byte order on FreeBSD 14, or when sending with those fields in network byte order on macOS Sonoma, you get `EINVAL` from the `sendto()` calls.  This can be verified not only experimentally, but also by comparing the «Raw IP Sockets» section of `ip(4)` for macOS and different versions of FreeBSD.

The BSD loopback encapsulation change on the other hand is on the libpcap receive path, and is only relevant to certain types of interfaces, including `utun` on macOS, which is used by at least some VPN software on macOS.  I have not ifdef'ed that to `__APPLE__`, as it seems to compile fine on Linux etc too.  The commit adding recv support for BSD loopback encapsulation is orthogonal to the rest of the changeset; sending in IP layer mode works without it.  Zmap would not be able to receive packets over utun interfaces tho, as used by at least some VPN software on macOS.  Byte order of the protocol field in the BSD loopback encapsulation header is currently irrelevant, as zmap never interprets it.

zakird (approved):

PR #778, opened 2024-02-11
Support IP layer mode on FreeBSD and merge Mac with BSD send code

droe:
- Add support for IP layer mode (`--iplayer` / `-X`) on FreeBSD, based on the macOS implementation
- Move `src/send-mac.h` code over to `src/send-bsd.h` and merge, as I think keeping the implementations separate is not worth the code duplication
- Updated sendmmsg comments re: FreeBSD

Originally I set out to port the sendmmsg code to BSD, but turns out that FreeBSD's sendmmsg is just a libc wrapper around sendmsg, which had escaped my attention when I initially looked into sendmmsg availability.  Which of course means sendmmsg would work, but without the perf benefit of less syscall overhead.  NetBSD on the other hand does have a proper sendmmsg syscall, but since I do not use NetBSD, I did not undertake the porting.

Tested on macOS Sonoma and FreeBSD 14.


src/send-bsd.h
```diff
@@ -23,23 +24,47 @@
...
+		// Early BSD's raw IP sockets required IP headers to have len
+		// and off fields in host byte order, as they were being byte
+		// swapped on the way out.  Getting byte order wrong would
+		// result in EINVAL from sendto(2) below.
+		// Most modern BSD systems have fixed this and removed the byte
+		// swapping on raw IP sockets, while some, notably macOS, still
+		// require header fields in host byte order.
+		// See ip(4) for details on byte order requirements of raw IP
+		// sockets.
+		if (retry_ct == 0) {
```
phillip-stephens, line 49: I'm not quite sure why this logic would be dependent on this being the first attempt to send this packet (`retry_ct == 0`). Can we add a comment to explain or remove this if it's not needed?

src/send-bsd.h
```diff
@@ -23,23 +24,47 @@
...
-
 
 int send_run_init(UNUSED sock_t sock)
 {
 	// Don't need to do anything on BSD-like variants
 	return EXIT_SUCCESS;
 }
 
-int send_packet(sock_t sock, void *buf, int len, UNUSED uint32_t idx)
+int send_packet(sock_t sock, void *buf, int len, UNUSED uint32_t retry_ct)
```
phillip-stephens, line 34:
```suggestion
int send_packet(sock_t sock, void *buf, int len, uint32_t retry_ct)
```
IMO we can remove this flag, since at least in some compilation cases this field _is_ used. WDYT?

phillip-stephens (commented):
Looks good overall, just two comments.
cc: @droe

src/send-bsd.h
droe: Added comment in 0c9316b — it's because the caller passes us the same buffer on retries, still byte-swapped.

src/send-bsd.h
droe: The purpose of the `UNUSED` is to silence the compiler warning in the compilation case(s) where the argument is _not_ used.  Keeping the `UNUSED` makes sense, unless we use the argument in _all_ compilation cases.

phillip-stephens: Fair! Sounds good, resolving.

PR #783, opened 2024-02-15
phillip/781: elevate debug log if can't get a socket on bsd to fatal

phillip-stephens:
Noticed in testing this [PR](https://github.com/zmap/zmap/pull/778) that if the user forgets `sudo`, there were a bunch of errors printed on Mac. @droe had the exact same setup but wasn't seeing this behavior, which is odd. Either way, I think this fix makes sense. If you can't get a socket, you can't send packets, so this should be a fatal error, IMO.

### DIFF
--- a/src/socket-bsd.c
+++ b/src/socket-bsd.c
@@ -36,8 +36,8 @@ sock_t get_socket(UNUSED uint32_t id)
 	if (zconf.send_ip_pkts) {
 		int fd = socket(PF_INET, SOCK_RAW, IPPROTO_RAW);
 		if (fd == -1) {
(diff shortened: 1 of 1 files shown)


droe:
Concur failure to create socket should lead to a fatal error, two comments:
- `CAP_NET_RAW` mentioned in the error string is a Linux thing and not applicable to macOS or any of the BSDs.
- There are four more failure paths like this one, so I'm thinking a more thorough fix would cover those too.

phillip-stephens:
@droe appreciate your help with this and lending your knowledge of Linux + BSD networking.

1. Thanks, fixed this!
2. Fair, I've addressed. LMK if it looks good to you!

droe:
lgtm, thanks — and thanks for bearing with my PRs!

zakird (approved):

zakird:
looks good, thank you both!!

PR #789, opened 2024-02-22
Add a suite of integration tests to Github Actions

phillip-stephens:
This PR adds a suite of integration tests that run as a Github Action
- [x] Add a suite of integration tests using `--dryrun` and validating output for different inputs and CLI flags
- [x] Add a single "real" scanning test against known "good" IPs: `1.1.1.1` and `8.8.8.8`.
- [x] Add more context to manpages/`--help` text explaining "target" for `-n`
- [x] Adds a Github Action that run these tests


phillip-stephens:
@zakird I added the tests you suggested, plus some others as well. The one thing I thought was useful but wanted your thoughts was the single "real" scan in `test_scan.py` that scans `1.1.1.1` and `8.8.8.8`. Thought this was valuable, and assuming these IPs don't cease being active as DNS resolvers, should continue to work as designed. But open to suggestions if you disagree.

zakird (approved):

PR #790, opened 2024-02-24
Use stdbool instead of typedef uint8_t bool in dns probe module

droe:
Convert dns probe module to use proper stdbool booleans instead of uint8_t.

Paves the way for using stdbool in headers that are included by `src/probe_modules/module_dns.c`, removes the need for undefining bool for NetBSD, and is more consistent with existing other use of `bool` in ZMap's codebase.


phillip-stephens (approved):
LGTM, thanks for the improvement!

PR #791, opened 2024-02-24
Fix two crashes in dns probe module args handling

droe:
Fix two crashes in the DNS probe module's argument handling:

- Null pointer dereference on `--probe-module=dns` without `--probe-args`
- Null pointer dereference on `--probe-module=dns` with `--probe-args=""`


phillip-stephens (approved):
LGTM, thanks for catching this @droe !

PR #792, opened 2024-02-24
Fixed bug in the shard initialization w/ multi-port

phillip-stephens:
While working on the integration test PR, I noticed my tests around multiple ports were intermittently failing. After some investigation, I believe this is the fix. This just adds the check already present in `shard_get_next_target` [here](https://github.com/zmap/zmap/blob/main/src/shard.c#L185) to the `shard_roll_to_valid`.

My tests pass after this fix.


phillip-stephens:
@zakird I'm not too familiar with the sharding code, but this fix _seems_ reasonable, wdyt?

zakird (approved):

PR #798, opened 2024-03-01
Prevent fingerprinting ZMap by randomizing the IP ID

phillip-stephens:
The IP packet header field `IP Identification` field was hardcoded to `54321`.

This PR uses the aes fast random number generator to generate a unique IP ID for each packet. This should be faster than reading from `/dev/urandom` if we used `random.c`.

Video showing a packet capture from Wireshark and the unique IDs:

https://github.com/zmap/zmap/assets/23459798/19a5a40f-93f5-43ac-8389-48c011d3d672


examples/probe-modules/module_tcp_cisco_backdoor.c
```diff
@@ -64,7 +65,7 @@ static int synscan_init_perthread(void *buf, macaddr_t *src, macaddr_t *gw,
 
 static int synscan_make_packet(void *buf, UNUSED size_t *buf_len,
 			       ipaddr_n_t src_ip, ipaddr_n_t dst_ip, uint8_t ttl,
-			       uint32_t *validation, int probe_num,
+			       uint32_t *validation, int probe_num, aesrand_t *aes,
```
zakird, line 68: I think we need to make sure here that we're not passing in the same entropy that's passed in from the `uint32_t *validation` parameter. It might make sense to pass in IPID directly too since that's more tightly typed.

phillip-stephens: Fair, I've moved to only generating the `ip_id` in `send.c` [here](https://github.com/zmap/zmap/pull/798/files#diff-33c4faff4dd3631a679cba7327bb63d614db66b2d3d29899ecb9bf30462a9085R233) and passing it directly in.

Also, I changed to using `random_bytes` which reads from `/dev/urandom`, should be completely distinct from any other RNG in our code.

examples/probe-modules/module_tcp_cisco_backdoor.c
```diff
@@ -64,7 +65,7 @@ static int synscan_init_perthread(void *buf, macaddr_t *src, macaddr_t *gw,
 
 static int synscan_make_packet(void *buf, UNUSED size_t *buf_len,
 			       ipaddr_n_t src_ip, ipaddr_n_t dst_ip, uint8_t ttl,
-			       uint32_t *validation, int probe_num,
+			       uint32_t *validation, int probe_num, u_short ip_id,
```
zakird, line 68: can you use the actual sized ones, not short/long, e.g., `uint32_t` here?

src/send.c
```diff
@@ -376,7 +381,7 @@ int send_run(sock_t st, shard_t *s)
 			zconf.probe_module->make_packet(
 			    buf, &length, src_ip, current_ip,
 			    htons(current_port), ttl, validation, i,
-			    probe_data);
+			    aesrand_getword(aes_rand_gen), probe_data);
```
zakird, line 384: I think that this will be the same for every packet that goes out, if I'm reading this code correctly, not random for each packet, which is what you want. I think that you can just use the last 4 bytes of validation here since probe modules generally only use the first few bytes.

zakird: Actually I take that back, I just looked at the code, and it does look like it works as you expected. My apologies.

zakird (approved):
Looks good, can you just swap from `u_short` to `uint32_t` please?

examples/probe-modules/module_tcp_cisco_backdoor.c
phillip-stephens: Yep, good call, sounds safer

src/send.c
```diff
@@ -376,6 +381,7 @@ int send_run(sock_t st, shard_t *s)
 			zconf.probe_module->make_packet(
 			    buf, &length, src_ip, current_ip,
 			    htons(current_port), ttl, validation, i,
+			    (uint16_t)aesrand_getword(aes_rand_gen),
```
droe, line 384: This does an additional extra full blown AES block encryption for every single packet, in order to produce 16 bits of IP ID.  In my quick perf testing on a 10GbE box with 8 cores and netmap, this change reduces send rate from 10.44 Mp/s to 8.80 Mp/s.

Would a more lightweight PRNG make more sense here, think cyclic group like for target permutation, instead of the AES-based PRNG?

phillip-stephens: Hmm, that sounds like a significant perf. hit, thanks for running those numbers! Let me look at grabbing the last 16 bits from permutation like @zakird mentioned, I'll double-check no probe-module uses those.

droe: fwiw, 46fac21 resolves the perf hit for me.

examples/probe-modules/module_tcp_cisco_backdoor.c
```diff
@@ -17,6 +17,7 @@
 
 #include "../../lib/includes.h"
 #include "../fieldset.h"
+#include "aesrand.h"
```
zakird, line 20: I believe this can be removed.

src/send.c
```diff
@@ -228,6 +228,11 @@ int send_run(sock_t st, shard_t *s)
 		zconf.probe_module->thread_initialize(
 		    buf, zconf.hw_mac, zconf.gw_mac, &probe_data);
 	}
+	// re-seeding the random number generator from /dev/urandom to ensure
+	// it is seeded different from any other RNG in ZMap
+	uint64_t ip_id_seed;
+	random_bytes(&ip_id_seed, sizeof(ip_id_seed));
+	aesrand_t* aes_rand_gen = aesrand_init_from_seed(ip_id_seed);
```
zakird, line 235: I believe this code can be removed now

zakird (changes requested):
New approach looks reasonable, but looks like there's some leftover code that needs to be removed.

phillip-stephens:
@zakird Yep, sorry about that, forgot to cleanup after switching the approach. Re-requesting review.

zakird (approved):

PR #803, opened 2024-03-03
AES hardware acceleration using AES-NI and ARMv8 CE

droe:
Introduce aes128 library code that uses hardware accelerated AES on architectures where it's supported, and will fall back to the existing software AES (Rijndael) if built without hardware acceleration, or if hardware acceleration is not supported by the CPU at run-time.

For now hardware acceleration is opt-in and requires `-DWITH_AES_HW` cmake argument to build. Acceleration implemented for x86_64 (AES-NI) and arm64 (ARMv8 CE) using compiler intrinsics, no assembly.

Adopt new aes128 abstraction in validate and aesrand.  Invoke quick selftest at ZMap start, to ensure correct operation regardless of which AES implementation is being used.

On hardware configurations where scan rate is limited by CPU, hardware-accelerating AES can substantially improve scan rate, because packet validation as currently implemented requires an AES block encryption for every packet sent and received.  I'm seeing a scan rate improvement from 10.28 Mp/s to 12.13 Mp/s with netmap on 10 GbE when using AES-NI instead of software AES.

Tested on macOS Sonoma (ARMv8 CE, clang), FreeBSD 14.0 (both AES-NI and ARMv8 CE, clang), and Ubuntu 23.10 (ARMv8 CE, GCC).

### DIFF
--- a/lib/aes128.c
+++ b/lib/aes128.c
@@ -1,0 +1,346 @@
+/*
+ * This file copyright 2024, Daniel Roethlisberger <daniel@roe.ch>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "aes128.h"
+
+#include "../lib/rijndael-alg-fst.h"
+#include "../lib/logger.h"
+
+#include <stdbool.h>
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+#include <pthread.h>
+
+
+#define AES128_ROUNDS 10
+#define BITS_PER_BYTE 8
+
+
+#ifdef AES_HW
+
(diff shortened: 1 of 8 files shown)


lib/aes128.c
```diff
@@ -0,0 +1,346 @@
+/*
```
phillip-stephens, line 1: So unfortunately we'll need to keep the same OSS license/copyright as on the other files (with Copyright belonging to the University of Michigan), without specific contributors. 
Long-term the goal is to setup a non-profit to transfer copyright ownership, and any change like this would require the approval of every contributor if we allowed copyrights for all contributors.

phillip-stephens (changes requested):
I still need to go through this more thoroughly but overall this looks great and a much needed/appreciated improvement. Especially with your other improvements on increasing send rate, CPU bottlenecking becomes more possible.

Do need to request changes to the copyright, more details in my comment, but thank you overall!

lib/aes128.c
droe: Makes sense, fixed in 1e104d4.

droe:
To be fair, while this was fun and turned out more presentable than I initially feared, it would probably be better to replace the hw acceleration bits (as well as the software Rijndael routines) with OpenSSL at some point, for an always-on solution.  I just could not bring myself to deal with the plethora of portability challenges that come with using the various flavours of OpenSSL.

zakird:
Yeah, I don't really want to have to deal with OpenSSL in all honesty. I think given that this is a performance optimization that you'd probably only really care about for super high send rates, it's fine to have it only work on x86_64 and arm64 is pretty reasonable.

phillip-stephens (approved):
LGTM, thank you!

PR #807, opened 2024-03-04
netmap: Reply to ARP requests from gateway for scan source IPs

droe:
In netmap mode, the OS network stack never gets to see incoming packets unless we explicitly forward them to the host rings; hence the kernel will not be responding to ARP requests.  To remove the need for static ARP entries on the gateway, respond to ARP requests from the gateway for any of the source IPs of the scan.

The recv thread can now submit packets for sending to send thread 0 via a zqueue.  This is used to respond to ARP requests that arrive after having let loose the send threads.  While still waiting for end-to-end connectivity with `--netmap-wait-ping`, ARP requests are responded to by sending directly from the recv thread, just like the ICMP Echo requests.

I did not see any significant netmap perf change from this at 10 GbE.

Tested on both FreeBSD and Linux.  Verified to correctly respond both to arping, as well as to real ARP by deleting ARP entries from the gateway's ARP table and observing the ARP table getting repopulated as well as packets on the wire.

### DIFF
--- a/src/recv-netmap.c
+++ b/src/recv-netmap.c
@@ -23,6 +23,7 @@
 #include "../lib/logger.h"
 
 #include <net/netmap_user.h>
+#include <net/if_arp.h>
 #include <sys/mman.h>
 #include <sys/ioctl.h>
 #include <fcntl.h>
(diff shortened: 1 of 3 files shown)


src/recv-netmap.c
```diff
@@ -33,41 +34,115 @@
...
+	submit_batch_internal(batch); // consumes batch
+}
+
+#define ARP_ETHER_INET_PKT_LEN (sizeof(struct ether_header) + sizeof(struct arphdr) + 2 * ETHER_ADDR_LEN + 2 * sizeof(uint32_t))
+#define x_ar_sha(ap) ((uint8_t *)((ap) + 1))
+#define x_ar_spa(ap) (((uint8_t *)((ap) + 1)) + ETHER_ADDR_LEN)
+#define x_ar_tha(ap) (((uint8_t *)((ap) + 1)) + ETHER_ADDR_LEN + sizeof(uint32_t))
+#define x_ar_tpa(ap) (((uint8_t *)((ap) + 1)) + 2 * ETHER_ADDR_LEN + sizeof(uint32_t))
+
+static size_t
```
phillip-stephens, line 83: Could we add the context you wrote in the PR description here? Think it might be great for future users of this file to know _why_ we need to specially handle ARP with `netmap`

```In netmap mode, the OS network stack never gets to see incoming packets unless we explicitly forward them to the host rings; hence the kernel will not be responding to ARP requests. To remove the need for static ARP entries on the gateway, respond to ARP requests from the gateway for any of the source IPs of the scan.```

phillip-stephens (changes requested):
Reviewed the code and looks great and make sense to me. If we can just add some context around the ARP replies, I think this LGTM!

src/recv-netmap.c
droe: Good call, addressed in f844a78.  This is indeed crucial info for future readers of this code (including future me).

phillip-stephens (approved):
LGTM, thanks @droe !!!

PR #812, opened 2024-03-06
Remove the PF_RING Code

phillip-stephens:
PF_RING is no longer actively maintained and has been supplanted with netmap in the ZMap codebase thanks to @droe. This PR removes the PF_RING codebase.


zakird:
Have we been able to verify that netmap can send at 10gbE?

PR #818, opened 2024-03-10
+2.9% perf: Avoid copying in send core, further batch memory optimisation

droe:
- Avoid the extra memcpy in the send core by cleanly adapting the probe module interface to the introduction of batching (splitting thread init into thread init and packet preparation, allowing the latter to be called on each buffer in a batch)
- Remove almost unused `ip` field from `batch->packets[]` and align the packet data such that the IP headers start at a 32 bit boundary, speeding up 32 bit header field access

The former results in a ~2% and the latter a further ~0.9% improvement to send rate at 10 GbE on lower-end hardware, while (arguably) improving code structure, or at least not making it worse.

Tested on macOS Sonoma, FreeBSD 14 and Ubuntu 23.10; smoke tested various probe modules.

### DIFF
--- a/src/probe_modules/module_bacnet.c
+++ b/src/probe_modules/module_bacnet.c
@@ -37,7 +37,16 @@ static inline uint8_t get_invoke_id(uint32_t *validation)
 	return (uint8_t)((validation[1] >> 24) & 0xFF);
 }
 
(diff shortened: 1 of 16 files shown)


droe:
I broke something, let me come back with this at a later time.

PR #820, opened 2024-03-13
Less copying, more batch optimisation

droe:
- Switch from writing each packet in a single buffer, and copying that buffer into the batch, to directly letting the probe modules write into the batch.  To that end, split probe module's `thread_initialize` callback into `thread_initialize` and `prepare_packet`, allowing the latter to be called on each buffer in a batch.
- Remove the almost unused `ip` field from `batch->packets[]`, as it was only used on an error path of BSD's send code, and the IP can also be read from the packet data.  While removing the `ip` field, align the packet data in batch such that the IP headers start at a 32-bit boundary, speeding up 32-bit header field access.

These combined get me ~ 2.9 % send rate improvement, two thirds of which coming from the first change.  Again, I realize some or all of this may be arguable.

The change in probe module interface could also be a step towards potential future work of letting probe modules write directly into mapped NIC memory in netmap mode (tho I suspect that might be more intrusive than it's worth).

This is a revised version of #818, now without breaking `--dryrun`.  Tested on macOS Sonoma, FreeBSD 14 and Ubuntu 23.10; smoke tested various probe modules; `test-shard.sh` and `test_big_group.sh` succeed.

### DIFF
--- a/src/probe_modules/module_bacnet.c
+++ b/src/probe_modules/module_bacnet.c
@@ -37,7 +37,16 @@ static inline uint8_t get_invoke_id(uint32_t *validation)
 	return (uint8_t)((validation[1] >> 24) & 0xFF);
 }
 
-int bacnet_init_perthread(void *buf, macaddr_t *src, macaddr_t *gw, void **arg)
+int bacnet_init_perthread(void **arg)
+{
+	uint32_t seed = aesrand_getword(zconf.aes);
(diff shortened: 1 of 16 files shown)


zakird (approved):
Looks sane to me. @phillip-stephens can you take a deeper pass?

phillip-stephens (approved):
LGTM! I wasn't able to test the `--iplayer` option but that's due to issue #821 , so I'm good to merge.

PR #838, opened 2024-03-23
Clang Format

zakird:
This PR runs `clang-format` using the repository rules, except for I've disabled 80 character line counts because blindly running it makes our code much less readable in many places. There is no change in behavior.


phillip-stephens (approved):
Sounds good to me!

Ale158-star:
Awesome

PR #848, opened 2024-04-11
Use libpcap timestamp information instead of calling gettimeofday() in receive thread

ogasser:
For every packet that ZMap receives it calls `gettimeofday()` in the receive thread to write the receive time to the output file. Instead, the already present timestamp in the PCAP header can be used which should save CPU cycles and lead to more accurate receive times.


zakird (approved):
Makes sense to me! Thanks for submitting the PR!!

PR #851, opened 2024-04-18
Increase sending slow_rate threshold to 200 

mpfaehler:
Thanks for maintaining the awesome ZMap.
I'd like to perform scans with a rate of 100pps. I've observed that ZMap consumes a lot of CPU in that case. However, if I increase the slow rate threshold in send.c (to maybe 200), everything is fine - would you like to merge this to main? 

Of course, don't merge if there is a profound reason of '50' being the current threshold.

### DIFF
--- a/src/send.c
+++ b/src/send.c
@@ -254,7 +254,7 @@ int send_run(sock_t st, shard_t *s)
 	double send_rate =
 	    (double)zconf.rate /
 	    ((double)zconf.senders * zconf.packet_streams);
-	const double slow_rate = 500; // packets per seconds per thread
+	const double slow_rate = 1000; // packets per seconds per thread
 	// at which it uses the slow methods
 	long nsec_per_sec = 1000 * 1000 * 1000;
 	long long sleep_time = nsec_per_sec;
--- a/src/send.c
+++ b/src/send.c
@@ -254,7 +254,7 @@ int send_run(sock_t st, shard_t *s)
 	double send_rate =
 	    (double)zconf.rate /
 	    ((double)zconf.senders * zconf.packet_streams);
-	const double slow_rate = 50; // packets per seconds per thread
+	const double slow_rate = 200; // packets per seconds per thread
 	// at which it uses the slow methods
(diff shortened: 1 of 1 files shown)


zakird:
The 50 is pretty arbitrary. What it comes down to is what can we accurately do given that `sleep` and friends are incredibly inaccurate in practice and it was hard to guarantee the specified send rate. @phillip-stephens can you test this on some of our boxes and see what rate we can run up to without our send rates going haywire? It might be that we can increase even past 200.

phillip-stephens:
It seems we have 2 approaches to ensuring the desired send rate, busy-waiting and sleep. Busy-waiting is accurate but all sending cores will stay maxed at 100%. Sleep frees up the core but is p. inaccurate esp. as you try to sleep for smaller intervals as you would with higher send rates.

On a machine with 20 cores, testing with both `-T 1` and `-T 4` (one and 4 sending threads), I saw we could push up to `slow_rate = 5000` with only ~1% inaccuracy in actual send rate when compared to expected send rate. 10k seemed to have some weird behavior where it was _reasonably accurate_sometimes, and 50% off other times.

However, it seems online `Sleep()` can become far less accurate as contention on the host increases. I realized my machine is likely a best case scenario with tons of cores and no other major processes going on. So since it's hardware-dependent, should we both:
1. increase the `slow_rate` to 5k  AND
2. add an optional override flag `--use-busy-wait` or `--use-sleep` to override this behavior if a user has a specific use-case/hardware scenario?

phillip-stephens:
@dadrian just tagging you to see if you have any thoughts here too 😄

dadrian:
I was going to guess that the sleep approach would work well until someone in the 1000-10000 range, because sleep usually thought of in milliseconds or fractional milliseconds. So increasing the threshold to 5000 sounds good to me, as do the flags.

zakird:
I'm good with increasing to 1000 or 5000.

zakird:
I'm going to merge this at 1000 and then @phillip-stephens can submit a new PR for the CLI flag.

zakird (approved):

PR #859, opened 2024-04-26
Fix unknown probe-args guidance for none -> smallest-probes change

droe:
Update guidance in error message to list `smallest-probes` instead of `none` which no longer exists.

#855 changed `--probe-args=none` to `--probe-args=smallest-probes`, but the error message for unknown `--probe-args` arguments still lists `none` as a valid argument (which is especially confusing when using `--probe-args=none` results in guidance suggesting to use `none`).


droe:
Integration test failure looks unrelated to change under test.

zakird (approved):

PR #886, opened 2024-07-02
the `--max-targets` flag with a percent wasn't respecting multi-ports

phillip-stephens:
## Before Fix - The `--max-targets` flag wasn't behaving as expected for a percent of the address space with multiple ports.

This manifested both in debug messages and in the ETA remaining un-changed when you use more than 1 port with this flag.

Single Port, 1% of address space
```
sudo ./src/zmap -B 100M -o "/dev/null"  --verbosity=5 -n 1% -p 80
...
Jul 02 23:08:51.018 [DEBUG] iterator: max targets is 42949672
...
 0:05 2% (5m16s left); send: 691206 139 Kp/s (137 Kp/s avg); recv: 8975 2.03 Kp/s (1.78 Kp/s avg); drops: 0 p/s (0 p/s avg); hitrate: 1.30%
```
2 Ports, 1 % of address space
```
sudo ./src/zmap -B 100M -o "/dev/null"  --verbosity=5 -n 1% -p 80-81
...
Jul 02 23:09:35.911 [DEBUG] iterator: max targets is 42949672
0:05 2% (5m21s left); send: 683403 135 Kp/s (135 Kp/s avg); recv: 4851 987 p/s (960 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.71%
```

## After Fix, note that `max targets` is correct _and_ the ETA is as expected

```
~/zmap-dev on  phillip/max-targets-percent-help! ⌚ 23:10:53
$ sudo ./src/zmap -B 100M -o "/dev/null"  --verbosity=5 -n 1% -p 80-81
...
Jul 02 23:11:05.208 [DEBUG] iterator: max targets is 85899345
...
 0:05 1% (10m left); send: 689020 137 Kp/s (136 Kp/s avg); recv: 4770 926 p/s (942 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.69%
```

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -144,8 +144,9 @@ void fprintw(FILE *f, const char *s, size_t w)
 	free(news);
 }
 
-uint32_t parse_max_hosts(char *max_targets)
+uint32_t parse_max_hosts(char *max_targets, int port_count)
 {
+	assert(port_count > 0);
 	char *end;
 	errno = 0;
 	double v = strtod(max_targets, &end);
@@ -153,8 +154,8 @@ uint32_t parse_max_hosts(char *max_targets)
 		log_fatal("argparse", "can't convert max-targets to a number");
 	}
 	if (end[0] == '%' && end[1] == '\0') {
-		// treat as percentage
-		v = v * ((unsigned long long int)1 << 32) / 100.;
+		// treat as percentage of the search space, so percent of 2^32 * num_ports
+		v = v * (((unsigned long long int)1 << 32) * port_count) / 100.;
 	} else if (end[0] != '\0') {
 		log_fatal("eargparse", "extra characters after max-targets");
 	}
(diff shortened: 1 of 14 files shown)


lib/util.c
```diff
@@ -144,24 +144,27 @@ void fprintw(FILE *f, const char *s, size_t w)
...
 	errno = 0;
 	double v = strtod(max_targets, &end);
 	if (end == max_targets || errno != 0) {
 		log_fatal("argparse", "can't convert max-targets to a number");
 	}
 	if (end[0] == '%' && end[1] == '\0') {
-		// treat as percentage
-		v = v * ((unsigned long long int)1 << 32) / 100.;
+		// treat as percentage of the search space, so percent of 2^32 * num_ports
+		v = v * (((unsigned long long int)1 << 32) * port_count) / 100.;
```
zakird, line 160: I would define this using something like `uint64_t` to be explicit about what you want here. We do that other places within the code base.

zakird (approved):

PR #888, opened 2024-07-09
Bump docker base image and relevant packages

phillip-stephens:
When this Dockerfile was last edited, `Ubuntu 22.04` hadn't been released yet. This PR bumps the base image and package versions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 #     docker run -it --rm --net=host ghcr.io/zmap/zmap <zmap args>
 ####
 
-FROM ubuntu:20.04 as builder
+FROM arm64v8/ubuntu:24.04 as builder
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -44,16 +44,16 @@ RUN cd /usr/local/src \
...
 LABEL org.opencontainers.image.source="https://github.com/zmap/zmap"
 
 RUN apt-get update \
     && apt-get install -y \
     libpcap0.8 \
-    libjson-c4 \
+    libjson-c5 \
     libjudydebian1 \
-    libhiredis0.14 \
+    libhiredis1.1.0 \
```
zakird, line 56: I don't think we should need Redis anymore given that we removed this output module a while back.

zakird (approved):
Approved modulo that I think you can remove a dependency.

Issue #736, opened 2023-10-23
Validation of destination port fails


davideandres95:
**Describe the bug**
Validation of destination port fails if `validation[1] % num_ports` is greater than `(validation[1] + zconf.packet_streams - 1) % num_ports` at:

https://github.com/zmap/zmap/blob/146ded1a1949220c815302df26052dbc9bfa12a4/src/probe_modules/packet.h#L147C1-L159

This can happen when the group is completely cycled.

**CLI Arguments**

```
zmap --probe-module=tcp_synscan --output-module=json --output-file=res.json --output-fields=success,ttl,saddr,daddr,repeat,timestamp_ts,timestamp_us,classification --target-port=80 --source-port=40002-40005 --gateway-mac=fa:76:a0:08:60:5f --allowlist-file=whitelist.conf --probes=3 --output-filter= --interface=def-red --bandwidth=100M --rate=0 --sender-threads=1 --source-ip=10.0.1.1 --verbosity 4
```

**Example Target IP**
I am running netcat at my local target (10.0.3.252) port 80.

**Expected behavior**
Validation should pass

**Environment:**
 - OS: Ubuntu Server 16.04
 - Version: Github HEAD

zakird:
Just writing down some notes, validated that this is indeed broken and test here:

```c
// Returns 0 if dst_port is outside the expected valid range, non-zero otherwise
int check_dst_port(uint16_t port, int num_ports,
				 uint32_t *validation)
{
	if (port > SOURCE_PORT_LAST || port < SOURCE_PORT_FIRST) {
		return 0;
	}
	int32_t to_validate = port - SOURCE_PORT_FIRST;
	int32_t min = validation[1] % num_ports;
	int32_t max = (validation[1] + PACKET_STREAMS  - 1) % num_ports;

	return (((max - min) % num_ports) >= ((to_validate - min) % num_ports));
}

uint16_t get_src_port(int num_ports, int probe_num, uint32_t *validation)
{
	return SOURCE_PORT_FIRST +
	       ((validation[1] + probe_num) % num_ports);
}


//uint16_t ports[] = {80, 443, 50000};

int main(void) {
	srand(time(NULL));
	for (int t=0; t < TRIALS; t++) {
		printf("Trial %i\n", t);
		uint32_t validation[4];
		validation[0] = rand();
		validation[1] = rand();
		validation[2] = rand();
		validation[3] = rand();
		for (int i=0; i< PACKET_STREAMS; i++) {
[...]

PR #891, opened 2024-07-16
Fix source port range size warning

Murgeye:
Currently, when giving any source port range (larger than a single port), zmap will return the message:

E.g.:

```
# zmap -p 443 -P 5 -s 50000-60000
[...]
[WARN] zmap: ZMap is configured to use a relatively small number of source ports (fewer than 10x the number of probe packets per target ip/port), which limits the entropy that ZMap has available for  validating responses. We recommend that you use a larger port range.
```

This is due to the inequality sign being flipped. Also see #736.


phillip-stephens (approved):
LGTM, thanks for this fix @Murgeye!!

PR #895, opened 2024-09-03
Handle upgrade path for blacklist to blocklist

### LINKED ISSUE #894: Error: zmap: unrecognized option '--blacklist-file=/etc/zmap/blacklist.conf'
**Describe the bug**
'blacklist' seems to be used internally somewhere. I see that this was changed to use 'blocklist'? Anyone having the same issue on a clean compile?

**CLI Arguments**
zmap --version
zmap -p 80
zmap <literally anything>

```
(zmap: unrecognized option '--blacklist-file=/etc/zmap/blacklist.conf')
```

phillip-stephens:
Closes #894 

## Description
If someone had installed zmap prior to changing the `blacklist` -> `blocklist`, they may have an existing `/etc/zmap/blacklist.conf` file. If this file had been changed or had any modifications in it, it will not be respected after installing the latest ZMap. The fix is to rename the existing file (if it exists) to the expected `etc/zmap/blocklist.conf`,

Additionally, (and what I believe caused #894), the contents of `zmap.conf` would have been 

```
blacklist-file "/etc/zmap/blacklist.conf"
```

If the `zmap.conf` file exists and contains that string, we'll replace it with:
```
blocklist-file "/etc/zmap/blocklist.conf"
```

Also fixed a typo.

### `main` when there's a `blacklist.conf` file

Notice how there are both a `blacklist.conf` and a `blocklist.conf`
```
276f9edbd3c2# mkdir /etc/zmap
276f9edbd3c2# echo "hello" > /etc/zmap/blacklist.conf
276f9edbd3c2# cmake . && make -j4 && make install
...
...
f13c372b071b# ls /etc/zmap
blacklist.conf  blocklist.conf  zmap.conf
f13c372b071b# cat /etc/zmap/blacklist.conf 
hello
f13c372b071b# cat /etc/zmap/blocklist.conf
# From IANA IPv4 Special-Purpose Address Registry
# http://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml
# Updated 2013-05-22

0.0.0.0/8           # RFC1122: "This host on this network"
10.0.0.0/8          # RFC1918: Private-Use
100.64.0.0/10       # RFC6598: Shared Address Space
127.0.0.0/8         # RFC1122: Loopback
169.254.0.0/16      # RFC3927: Link Local
...
```

### `Phillip/894` when there's a pre-existing `blacklist.conf` file
```
276f9edbd3c2# mkdir /etc/zmap
276f9edbd3c2# echo "hello" > /etc/zmap/blacklist.conf
276f9edbd3c2# cmake . && make -j4 && make install
...
67fc06685b98# ls /etc/zmap
blacklist.conf  blocklist.conf  zmap.conf
67fc06685b98# cat /etc/zmap/blacklist.conf 
hello
67fc06685b98# cat /etc/zmap/blocklist.conf
hello

```

### `Phillip/894` with a fresh install
```
80a30c9ecba2# cmake . && make -j4 && make install
...
80a30c9ecba2# ls /etc/zmap
blocklist.conf  zmap.conf
80a30c9ecba2# cat /etc/zmap/blocklist.conf
# From IANA IPv4 Special-Purpose Address Registry
# http://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml
# Updated 2013-05-22

0.0.0.0/8           # RFC1122: "This host on this network"
10.0.0.0/8          # RFC1918: Private-Use
100.64.0.0/10       # RFC6598: Shared Address Space
127.0.0.0/8         # RFC1122: Loopback
169.254.0.0/16      # RFC3927: Link Local
172.16.0.0/12       # RFC1918: Private-Use
...
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,6 +202,12 @@ foreach(EACH_CONF ${CONF_FILES})
     endif()
 endforeach()
 
+# Upgrade path for old conf files
+if(EXISTS "${CONFIG_DESTINATION}/blacklist.conf")
+    message(STATUS "Old configuration file detected at ${CONFIG_DESTINATION}/blacklist.conf, moving to ${CONFIG_DESTINATION}/blocklist.conf to match the new flag conventions")
+    file(RENAME "${CONFIG_DESTINATION}/blacklist.conf" "${CONFIG_DESTINATION}/blocklist.conf")
+    message(STATUS "blocklist.conf has been successfully moved.")
+
 # Allow Debian Packaging
 include(InstallRequiredSystemLibraries)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,6 +188,14 @@ endif()
 
 FILE(GLOB CONF_FILES "${PROJECT_SOURCE_DIR}/conf/*")
 
+# Upgrade path for old conf files - this is necessary for users upgrading from a version where we called the blocklist file blacklist.conf
+# Must occur before the install command
+if(EXISTS "${CONFIG_DESTINATION}/blacklist.conf")
+    message(STATUS "Old configuration file detected at ${CONFIG_DESTINATION}/blacklist.conf, moving to ${CONFIG_DESTINATION}/blocklist.conf to match the new flag conventions")
+    file(RENAME "${CONFIG_DESTINATION}/blacklist.conf" "${CONFIG_DESTINATION}/blocklist.conf")
+    message(STATUS "blocklist.conf has been successfully moved.")
+endif()
+
 message(STATUS "Default ZMap configuration file location is /etc/zmap")
 foreach(EACH_CONF ${CONF_FILES})
     get_filename_component(CONF_BASENAME ${EACH_CONF} NAME)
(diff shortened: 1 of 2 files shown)


zakird:
I'm a little bit hesitant to do this because it could unexpectedly break things for the user. What if we instead added a symlink?

phillip-stephens:
That seems fine and perhaps less intrusive. We still have the issue with `zmap.conf` that I just realized and mention in the now updated description. Not sure if we should just leave that as an exercise to the user (could always just do a fresh install) but I added code that should handle it. Tested with a pre-existing `zmap.conf` and installed latest zmap.
cc: @zakird

zakird:
Is that an issue if the symlink is there?

phillip-stephens:
Yeah, probably should have expanded a bit more in the PR description. The symlink solves the issue with having a `blacklist.conf`, we can just create a symlink to `blocklist.conf -> blacklist.conf`. BUT, there's a 2nd issue with the `zmap.conf` config file.

We have the "standard options" run with most `zmap` invocations stored in `/etc/zmap/zmap.conf`.

The current version looks like:
```
...
### Blocklist file to use. We encourage you to exclude
### RFC1918, IANA reserved, and multicast networks,
### in addition to those who have opted out of your
### network scans.
blocklist-file "/etc/zmap/blocklist.conf"

...
```

If the `.../zmap.conf` file already exists, it won't be re-created by default. See [here](https://github.com/zmap/zmap/blob/main/CMakeLists.txt#L195).

So I believe that in #894, the user's `zmap.conf` file had the old line: `blacklist-file "/etc/zmap/blacklist.conf"`. That would have caused the error they saw. I think the only fix would be to either _always_ overwrite the old file if it exists, leave it to the user to fix, or replace that string if it exists.

zakird (approved):

PR #912, opened 2024-12-02
Fix multi-port scans aborting early

phillip-stephens:
This fix lets multi-port scans proceed past 2^32 = 4,294,967,296. 

However, I've seen some weird behavior:

1. ETA of scan reaching 100% around the mid-way point of scanning
2. Scanning far too many targets given the CLI inputs

Going to investigate further and add some repeatable integration tests for full IPv4 multi-port scans. I don't want to merge this until I've at least opened issues/verified the issues I saw over the weekend so `main` isn't in a bad state with no documentation.

### DIFF
--- a/src/send.c
+++ b/src/send.c
@@ -413,11 +413,10 @@ int send_run(sock_t st, shard_t *s)
 			batch->packets[batch->len].len = (uint32_t)length;
 
 			if (zconf.dryrun) {
-				// TODO Remove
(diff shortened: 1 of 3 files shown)


zakird (approved):

phillip-stephens:
Closing since pushed extra integration test changes to this branch, reopened this fix on #914

PR #921, opened 2025-01-03
Fix leaks

rex4539:


phillip-stephens (approved):
Awesome, thanks for this catch @rex4539 !

PR #931, opened 2025-03-05
Implement NTP_TIMESTAMP (RFC5905) template field

svbatalov:
This PR implements `NTP_TIMESTAMP` template field that produces a 64 bit NTP timestamp value in accordance with [RFC5905](https://datatracker.ietf.org/doc/html/rfc5905).

Example usage:

```sh
$ echo -n '${HEX=d30004fa000100000001000000000000000000000000000000000000000000000000000000000000}${NTP_TIMESTAMP}' > ntp_time.tpl

$ zmap -qv0 -M udp -p 123 --probe-args=template:ntp_time.tpl --output-module=json --output-fields=saddr,data 216.239.35.0 | jq .
{
  "saddr": "216.239.35.0",
  "data": "140104ec0000000000000006474f4f47eb72653504513460eb726534b875a31aeb72653504513461eb72653504513463"
}
```

Field description:
```sh
$ zmap -qv0 -M udp -p 123 --probe-args=template-fields
...
NTP_TIMESTAMP: 64 bit NTP timestamp in network byte order (RFC5905)
```


zakird (approved):
Looks great — thank you!!

PR #954, opened 2025-11-07
README update - include mention of scanning ethics and example

phillip-stephens:
Have held a number of conversations with people over the past few weeks at IMC and elsewhere and a common thread is that:

1. Mention front-and-center in the README about ethical implications of scanning
2. Include an example showing how easy it is to use

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,14 +3,6 @@ ZMap: The Internet Scanner
 
 ![Build Status](https://github.com/zmap/zmap/actions/workflows/cmake.yml/badge.svg)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,6 +3,14 @@ ZMap: The Internet Scanner
 
 ![Build Status](https://github.com/zmap/zmap/actions/workflows/cmake.yml/badge.svg)
 
+> [!CAUTION]
```
zakird: I would move this after the description of what ZMap is and before Using ZMap.

PR #970, opened 2026-05-11
Adds a RTT measurement to the TCP SYN probe

### LINKED ISSUE #902: [Feature Request] TCP Ping Feature
While we do have the `icmp_echo_time` module, it could be useful to add a ping to TCP as well as an optional output field. 

With the `seq_no` field we have 32 bits, so I'm thinking we can store 1/10 ms. precision timestamps in the `seq_no` as long as scans don't take longer than (2^32) / (10^4) / 60 / 60 = 119 hours.

Then on probe response reception, we can calculate the observed latency using the `ack_num`

New Output field name: `latency: time difference between when probe response was received and probe was sent, to 1/10 ms. precision`

phillip-stephens:
This PR adds a RTT measurement to the TCP SYN probe.

Building off @droe's PR (thank you!), this adds the following:

- New `--output-fields=rtt` that when selected, outputs RTT measurement _assuming_ `--probe-args=...rtt...` was enabled
- New `--probe-args` field `rtt` enables the RTT measurement. If `rtt` is included in `--output-fields` but not selected in `--probe-args`, we print a warning .
- Uses 24 bits of the TCP sequence number field to encode a timestamp of when the probe was sent, leaving 8 bits for the validation
- In the packet processing function, we use the timestamp on the pcap as the time the probe was received. This eliminates the delay between probe response received and packet processing function running, reducing measurement inaccuracy.

```shell
rtt             string: RTT in ms to one decimal place, max RTT reliably measured = ~28min
```

### RTT Accuracy
To ensure we were measuring the RTT properly, I sent a single ZMap probe to `1.1.1.1:80` and ran a pcap capture at the same time. You can see below that the timings between the two line up perfectly.

ZMap
```shell
sudo ./src/zmap --rate 1 -p 80 -c 1 --output-fields=saddr,sport,classification,success,rtt 1.1.1.1
...
saddr,sport,classification,success,rtt
...
1.1.1.1,80,synack,1,101.1
...
May 11 14:02:54.047 [INFO] zmap: completed
```

`tshark` Output
```shell
⋊> ~ tshark -i en0 -f "host 1.1.1.1 and tcp"                                                                    14:02:45
Capturing on 'Wi-Fi: en0'
    1   0.000000 192.168.0.204 → 1.1.1.1      TCP 66 58113 → 80 [SYN] Seq=0 Win=65535 Len=0 MSS=1460 WS=256 SACK_PERM
    2   0.101101      1.1.1.1 → 192.168.0.204 TCP 66 80 → 58113 [SYN, ACK] Seq=0 Ack=1 Win=65535 Len=0 MSS=1460 SACK_PERM WS=8192
    3   0.101423 192.168.0.204 → 1.1.1.1      TCP 54 58113 → 80 [RST] Seq=1 Win=0 Len=0
^C3 packets captured
``` 

### Loss of Validation Bits
ZMap encodes verification data into each probe to ensure that probe responses are from this particular zmap scan. By reducing the bits available for verification, I was worried that we'd see an increase in the successful packets received.

To test, I ran 5x back-to-back interleaved scans with ZMap `main` branch and this one. The results were that the successful probe response counts were not statistically different and so therefore, I don't think the reduction in bits available for verification is an issue.

```shell
$ python ttest.py
=== success_unique ===

Metric          zmap-main     zmap-rtt
--------------------------------------
n                       5            5
mean             660793.8     660775.2
stddev             1529.9        611.0
min                658878       659831
max                662802       661726
range                3924         1895

Welch's t-test:   t=0.0226,  p=0.9828
```

## Related Issues
Closes #902

### DIFF
--- a/src/probe_modules/module_tcp_synscan.c
+++ b/src/probe_modules/module_tcp_synscan.c
@@ -36,6 +36,59 @@ probe_module_t module_tcp_synscan;
 static uint16_t num_source_ports;
 static uint8_t os_for_tcp_options;
 
+#ifdef RTT
+#include <errno.h>
+
+#ifndef NSEC_PER_SEC
+#define NSEC_PER_SEC 1000000000
+#endif
+#ifndef NSEC_PER_MSEC
+#define NSEC_PER_MSEC 1000000
+#endif
+#ifndef MSEC_PER_SEC
+#define MSEC_PER_SEC 1000
+#endif
+
+static void
+timespec_get_monotonic(struct timespec *t)
+{
+	if (clock_gettime(CLOCK_MONOTONIC, t) == -1) {
+		log_fatal("recv-netmap", "Failed to obtain monotonic time: %d: %s", errno, strerror(errno));
+	}
+}
+
+static struct timespec
+timespec_diff(struct timespec const *t1, struct timespec const *t0)
+{
+	struct timespec diff = {
+	    .tv_sec = t1->tv_sec - t0->tv_sec,
+	    .tv_nsec = t1->tv_nsec - t0->tv_nsec,
+	};
+	if (diff.tv_nsec < 0) {
+		diff.tv_sec--;
+		diff.tv_nsec += NSEC_PER_SEC;
+	}
+	return diff;
+}
+
+static uint64_t
+timespec_to_ms64(struct timespec const *t)
+{
+	return (uint64_t)t->tv_sec * MSEC_PER_SEC + (uint64_t)t->tv_nsec / NSEC_PER_MSEC;
+}
+
+static uint64_t
+timespec_ticks_since(struct timespec const *t_ref)
+{
+	struct timespec t_now;
+	timespec_get_monotonic(&t_now);
+	struct timespec t_diff = timespec_diff(&t_now, t_ref);
+	return timespec_to_ms64(&t_diff);
+}
+
+static struct timespec t_begin;
+#endif // RTT
+
 static int synscan_global_initialize(struct state_conf *state)
 {
 	num_source_ports =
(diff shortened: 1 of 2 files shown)


src/probe_modules/module_tcp_synscan.c
```diff
@@ -51,35 +81,57 @@ static int synscan_global_initialize(struct state_conf *state)
 					 "defaulting to Windows-style TCP options. Windows-style TCP options offer the highest hit-rate with the least bytes per probe.");
 		state->probe_args = (char *)"windows";
 	}
-	if (strcmp(state->probe_args, "smallest-probes") == 0) {
+
```
zakird, line 92: It feels a little bit weird to me that if you did `--output-fields=*` you'd have a change in behavior. Output really isn't supposed to change operations. I think that this would probably be better off as a different probe module or a module configuration option to make that explicit.

phillip-stephens: Hmm, didn't think about that, we wouldn't want to break anyone who's expecting `--output-fields=*` to have a certain output. I could add it as a `--probe-args=rtt`? I actually had it like this before but then the user needs to request it "twice": `--probe-args=rtt --output-fields=rtt,...`.

If you think that's fine, I'll go ahead and add that. IMO I prefer the `--probe-args` approach to a second probe module that's identical except for also the RTT in the seq number, seems like a lot of code duplication.

zakird: I think that adding it as a probe argument makes sense. We could output as NULL and warn if it's not enabled via probe arguments but requested in the output fields.

droe:
Thank you for pushing this over the finish line, @phillip-stephens

src/probe_modules/module_tcp_synscan.c
phillip-stephens:
```shell
saddr,sport,rtt                                                                                                                                                                                                                 
May 12 12:33:33.172 [WARN] tcp_synscan: "rtt" is listed in --output-fields but RTT is not enabled; add "rtt" to --probe-args to enable it (e.g. --probe-args=rtt or --probe-args=windows,rtt)
...                                                                                                                                
 0:00 0%; send: 0 0 p/s (0 p/s avg); recv: 0 0 p/s (0 p/s avg); drops: 0 p/s (0 p/s avg); hitrate: 0.00%                                                                                                                        
1.1.1.1,80,
```

Done!

zakird (approved):

PR #974, opened 2026-05-20
[tcp_synscan] add optional JA4TS fingerprint of SYN-ACK responses

apowis:
## Summary

Adds opt-in JA4TS TCP fingerprinting to the `tcp_synscan` probe module. Pass `--probe-args=ja4ts` to emit a new `ja4ts` output field containing the FoxIO JA4TS fingerprint of each SYN-ACK response.

Format: `{window}_{tcp_option_kinds_in_order}_{mss}_{wscale_shift}` (decimal, dash-separated kinds, `00` sentinel for absent MSS/wscale). See https://medium.com/foxio/ja4t-tcp-fingerprinting-12fb7ce9cb5a for the fingerprint definition.

Combines with other probe-args: `--probe-args=windows,ja4ts`, `--probe-args=linux,rtt,ja4ts`, etc.

Example:

```
$ sudo zmap -p 53 --probe-args=windows,ja4ts -f saddr,ja4ts,classification 1.1.1.1 8.8.8.8
1.1.1.1,65535_2-4-8-1-3_1460_7,synack
8.8.8.8,65535_2-4-8-1-3_1460_7,synack
```

## Implementation

- Pure formatter `compute_ja4ts()` walks TCP options once, preserves option-kind order (including NOPs and EOL), captures raw MSS and wscale shift, and writes the JA4TS string.
- Gated behind a new `ja4ts` token in `--probe-args`, paralleling the existing `rtt` token. New `ja4ts` field is null for RST and ICMP responses, or when the feature isn't enabled.
- Warns if `ja4ts` appears in `--output-fields` but isn't enabled via `--probe-args`, matching the existing `rtt` behavior.
- `tcp_synackscan` is unchanged — JA4TS is a SYN-ACK fingerprint and has no application there.

## Tests

- **Unit tests** under `test/unit/` (built standalone via `make -C test/unit run`). Six cases cover: no options, MSS-only, Linux-like full options, wscale shift 0 (distinct from absent), malformed length (defensive stop), and EOL termination.
- **Integration test** under `test/integration-tests/test_ja4ts.py` runs `zmap --probe-args=ja4ts` against `1.1.1.1` / `8.8.8.8` and asserts the field is well-formed.

### DIFF
--- a/src/probe_modules/module_tcp_synscan.c
+++ b/src/probe_modules/module_tcp_synscan.c
@@ -118,6 +118,7 @@ probe_module_t module_tcp_synscan;
 static uint16_t num_source_ports;
 static uint8_t os_for_tcp_options;
 static bool rtt_enabled = false;
+static bool ja4ts_enabled = false;
 
 // RTT is encoded in the lower RTT_TIMESTAMP_BITS of the TCP SYN sequence number
 // (XOR'd with validation[0]). The upper RTT_VALIDATION_BITS are used for packet
@@ -177,6 +178,8 @@ static int synscan_global_initialize(struct state_conf *state)
 	while (token) {
 		if (strcmp(token, "rtt") == 0) {
 			rtt_enabled = true;
+		} else if (strcmp(token, "ja4ts") == 0) {
+			ja4ts_enabled = true;
 		} else if (strcmp(token, "smallest-probes") == 0) {
 			if (os_set) {
 				log_fatal("tcp_synscan", "multiple OS options specified in probe-args");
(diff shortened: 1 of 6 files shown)


phillip-stephens:
On first glance, this looks great. Appreciate following the other `--probe-arg` conventions.

I'm wondering though after reading that Medium article if this provides value without being able to scan with different combinations of TCP Options? From my understanding, the server will respond with a different set of TCP Options depending on what it receives in the `SYN`, so what workflows would benefit from having the `JA4TS` fingerprint from ZMap's default TCP options (a custom combination of options that isn't following any of the major OS's exactly)?

Like sharing these fingerprints would only provide value if the other scanner used the identical TCP Options as ZMap, no?

This might point more to my lack of experience in this area, so mostly just curious.

apowis:
@phillip-stephens great point about the `JA4TS` being different based off the initial request, I had forgotten about that detail.

I wasn't actually considering the sharing data, I was more thinking about being able to analyse data I know is from zmap.

Perhaps it would make sense to also expose the `JA4T` fingerprint too, that way when people do share scan results, more context is given? Would easy to do, plus we would only need to calculate the `JA4T` once per scan, not every response.

apowis:
@phillip-stephens - ok I got carried away there and added the JA4T in last commit. Let us know what you think? Feels appropriate to have these side by side in same PR.

phillip-stephens (approved):
LGTM thanks for the effort! 

I realized in testing that if a user runs: `zmap --probe-args"rtt,ja4ts,ja4t"`, they'd just receive IPs in output since the default value for `--output-fields=saddr`. I added a commit warning the user if they request one of these new behaviors but don't request them in output since that feels like a mistake. Know I missed doing it in `rtt` but felt like a good add. If that all looks good to you, feel free to merge when satisfied.

apowis:
Perfect, nice addition thanks. I'm away without laptop right now and unable to merge for some reason, so feel free to merge for me if you get a chance.

PR #977, opened 2026-07-07
WIP: Add Judy1 cacheset for dedup on 64-bit machines

### LINKED ISSUE #849: Use Judy1 on 64-bit Systems for Deduplication
Cachehash uses [JudyHS](https://judy.sourceforge.net/doc/JudyHS_3x.htm) on the backend for our deduplication. We originally used this because we wanted ZMap to remain usable on 32-bit systems. However, given that deduplication maps to have be fairly large empirically (>1M items), it would be helpful to use [Judy1](https://judy.sourceforge.net/doc/Judy1_3x.htm) when we are on a 64-bit system to save memory. This will require having a bitmap-style cachehash library that uses J1 on the backend.

jakobwsmnn:
Closes #849 

## Summary
Adds cacheset, which is a data structure with a Judy1 backend. It can be used for deduplication on 64-bit systems. On 32-bit systems, the existing cachehash is still being used because the needed data does not fit in 32 bits.

## Motivation
See #849 for motivation.

## Design Decisions
The current [cachehash](https://github.com/zmap/zmap/blob/main/lib/cachehash.h) implementation implements an LRU cache by using JudyHS. It maps each key to a node_t pointer. node_t is a doubly linked list which maintains recency order. This way an LRU cache can be easily implemented. 
To implement this with Judy1 we run into the following problem: Judy1 has no connection between the key and the value. Therefore implementing the LRU strategy isn't straightforward. 
This implementation uses FIFO eviction instead. Thus, for me, the question arises: is this fine in this scenario?

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -1,6 +1,7 @@
 SET(LIB_SOURCES
     blocklist.c
     cachehash.c
+    cacheset.c
     constraint.c
     logger.c
     pbm.c
--- a/lib/cacheset.c
+++ b/lib/cacheset.c
@@ -1,0 +1,84 @@
+/*
+ * CacheSet Copyright 2026 Jakob Wiesmann
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ */
+#include "cacheset.h"
+
+#include <stddef.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include <assert.h>
+
+#include <Judy.h>
+
+struct cacheset_s {
+	Pvoid_t judy;
+	uint64_t *ring;
+	size_t maxsize;
+	size_t currsize;
+	size_t head;
+};
+
+cacheset *cacheset_init(size_t maxitems)
+{
+	assert(maxitems > 0);
+	cacheset *retv = malloc(sizeof(cacheset));
+	assert(retv);
+	memset(retv, 0, sizeof(cacheset));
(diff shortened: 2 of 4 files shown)


jakobwsmnn:
I have following questions:
- Do you agree with my design decision to use FIFO instead?
- I currently don't know how to test this:
    - I've seen that there isn't a unit test for cachehash. Should I add one for cacheset?
    - Should I write an integration test instead?

zakird:
Generally we've used LRU because some hosts will send us back duplicate responses forever, which a FIFO queue will miss and they'll show up repeatedly later in the results.

jakobwsmnn:
@zakird Thank you for your quick response, so if LRU is required I currently don't know how to implement this with Judy1. I did a bit of research before implementing the cacheset and I came across JudyL which could potentially be used for this.

JudyL maps Word_t to Word_t. My idea is to use two JudyL array:
1. map a dedup key to a timestamp (monotonically increasing: could use an integer or something similar)
2. map the timestamp to the dedup key

In order to find the LRU entry, we can use JLF on the second JudyL array, which returns the smallest index. (the smallest index is the least-recently-used value)
To update an entry, we can delete the old timestamp and reinsert.

This would avoid the doubly linked list + malloc. So it's cheaper than cachehash. However it requires two JudyL arrays which store a Word_t per entry. 
So it's still more expensive than the Judy1 + ring buffer approach.

zakird:
I agree that we can't use Judy1 — I think that's an error and really we
want to be using JudyL. I think it's fine to keep the existing overall
structure with the ring buffer, we can just flip out the JudyHS calls for
JudyL calls when using a 64-bit system since word is large enough.

jakobwsmnn:
Okay, I will try to implement an version with JudyL.

